PR #148, opened 2025-03-07
chore: revert "Functions with literals and params should type resolve"

auxesis:
Reverts cipherstash/proxy#145

Merge commit on `main` has a [failing build](https://github.com/cipherstash/proxy/actions/runs/13713650806).


auxesis:
Cancelling because the issue has been fixed by cipherstash/encrypt-query-language#93, and the build on `main` is [passing again](https://github.com/cipherstash/proxy/actions/runs/13713650806).

PR #151, opened 2025-03-10
fix(mapper): handle multiple statements in simple query messages

tobyhede:
Parse simple query messages into Vec of Statement AST/s and encrypt each statement if required.
Adds new EncryptedText variant of Portal to distinguish Simple Protocol in the backend

### DIFF
--- a/packages/cipherstash-proxy/src/postgresql/frontend.rs
+++ b/packages/cipherstash-proxy/src/postgresql/frontend.rs
@@ -244,72 +244,96 @@ where
     async fn query_handler(&mut self, bytes: &BytesMut) -> Result<Option<BytesMut>, Error> {
         let mut query = Query::try_from(bytes)?;
 
-        let statement = self.parse_statement(&query.statement)?;
-        self.check_for_schema_change(&statement);
+        // Simple Query may contain many statements
+        let parsed_statements = self.parse_statements(&query.statement)?;
+        let mut transformed_statements = vec![];
 
-        if !eql_mapper::requires_type_check(&statement) {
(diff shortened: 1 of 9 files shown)


packages/cipherstash-proxy/src/postgresql/frontend.rs
```diff
@@ -244,72 +244,97 @@ where
...
-        self.check_for_schema_change(&statement);
+        // Simple Query may contain many statements
+        let parsed_statements = self.parse_statements(&query.statement)?;
+        let mut transformed_statements = vec![];
 
-        if !eql_mapper::requires_type_check(&statement) {
-            counter!(STATEMENTS_PASSTHROUGH_TOTAL).increment(1);
-            return Ok(None);
-        }
+        warn!(target: MAPPER,
```
yujiyokoo, line 251: Maybe this can be info or debug?

packages/cipherstash-proxy/src/postgresql/frontend.rs
```diff
@@ -244,72 +244,97 @@ where
...
+            client_id = self.context.client_id,
+            msg = "Query contains multiple statements",
+            statements = parsed_statements.len(),
```
yujiyokoo, line 254: Is there a check for this length being `> 1`, or is it implied some other way?

yujiyokoo (approved):
LGTM

packages/cipherstash-proxy/src/postgresql/frontend.rs
tobyhede: If parsed_statements is empty, we skip the loop and essentially return immediately as a Passthrough

PR #153, opened 2025-03-10
test(eql): Add null record to ORE operator tests

tobyhede:
Adds `null` record to ORE operator tests. Test behaviour is unchanged, null records should now be excluded from comparison, following existing PostgreSQL behaviour.

Fix is in EQL ore here https://github.com/cipherstash/encrypt-query-language/pull/94


yujiyokoo (approved):
The tests will pass with the EQL update, right?

tobyhede:
@yujiyokoo Yes, they should as soon as eql is latest

PR #160, opened 2025-03-12
feat(migrate): Import migrator code and add as proxy subcommand

tobyhede:
Migrator command assumes a running proxy, and uses existing config to make the connection. 
Streamlines the migrate options, as only table and source/destination columns are required. 

- [x] Pull in code from migrator app
- [x] Enable as proxy subcommand
- [x] Integration tests
  - command line exec via docker 
  - rust integration test via direct call

Example usage
```bash
cipherstash-proxy encrypt --table encrypted --columns plaintext=encrypted_text
```

### DIFF
--- a/packages/cipherstash-proxy/src/cli/migrate/mod.rs
+++ b/packages/cipherstash-proxy/src/cli/migrate/mod.rs
@@ -17,27 +17,34 @@ const LOCALHOST: &str = "127.0.0.1";
 #[derive(clap::Args, Clone, Debug)]
 #[command(version, about, long_about)]
 ///
-/// Encrypt data in table
+/// Encrypt one or more columns in table
+/// Requires a running and configured CipherStash Proxy instance.
 ///
 pub struct Migrate {
+    ///
+    /// Name of database table
+    ///
     #[arg(short, long)]
     table: String,
 
     ///
+    /// Source and destination columns as space-delimited key pairs `--columns source=destination`
+    ///
+    #[arg(short, required = true, long, num_args(1..), value_parser = parse_key_val::<String, String>)]
+    columns: Vec<(String, String)>,
+
+    ///
     /// Primary key column/s
-    /// Compound primary keys can be provided as a space delimted list: `--primary-key id user_id``
+    /// Compound primary keys can be provided as a space delimted list: `--primary-key id user_id`
     ///
     #[arg(short = 'k', long, num_args(1..), value_delimiter = ' ', default_values_t = vec![ID.to_string()])]
     primary_key: Vec<String>,
-
-    #[arg(short, long, num_args(1..), value_parser = parse_key_val::<String, String>)]
-    columns: Vec<(String, String)>,
 
     // Updates `batch_size` records at a time
     #[arg(short, long, default_value_t = 100)]
     batch_size: usize,
 
-    /// Run without update. Data is loaded, but updates are not performed.
+    /// Run without update. Data is fetched, but updates are not performed.
     #[arg(short, long, default_value_t = false)]
     dry_run: bool,
 
--- a/packages/cipherstash-proxy/src/cli/migrate/mod.rs
+++ b/packages/cipherstash-proxy/src/cli/migrate/mod.rs
@@ -20,6 +20,10 @@ pub struct Migrate {
     #[arg(short, long)]
     table: String,
 
+    ///
+    /// Primary key column/s
+    /// Compound primary keys can be provided as a space delimted list: `--primary-key id user_id``
+    ///
     #[arg(short = 'k', long, num_args(1..), value_delimiter = ' ', default_values_t = vec![ID.to_string()])]
     primary_key: Vec<String>,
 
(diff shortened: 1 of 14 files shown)


tests/tasks/test/integration/migrate.sh
```diff
@@ -0,0 +1,44 @@
...
+EOF
+
+docker exec -i proxy cipherstash-proxy encrypt --table encrypted --columns plaintext=encrypted_text  --verbose
+
+docker exec -i postgres${CONTAINER_SUFFIX} psql postgresql://cipherstash:password@proxy:6432/cipherstash <<-EOF
+SELECT * FROM encrypted;
+EOF
+
+set +e
+OUTPUT="$(docker exec -i postgres${CONTAINER_SUFFIX} psql 'postgresql://cipherstash:password@proxy:6432/cipherstash?sslmode=disable' --command 'SELECT id FROM encrypted WHERE encrypted_text IS NULL' 2>&1)"
```
yujiyokoo, line 29: Besides this non-null test, it would be great to also check the encrypted columns contain the expected values, perhaps with something like checking `WHERE encrypted_text = 'Three'` returns a row

packages/cipherstash-proxy/src/cli/migrate/mod.rs
```diff
@@ -0,0 +1,309 @@
...
+
+    tokio::spawn(async move {
+        if let Err(err) = connection.await {
+            error!("Connection error: {}", err);
+        }
+    });
+    Ok(client)
+}
+
+mod tests {
```
yujiyokoo, line 294: Can this be under `#[cfg(test)]`, unless it's needed outside of tests?

packages/cipherstash-proxy/src/cli/migrate/mod.rs
```diff
@@ -0,0 +1,309 @@
...
+
+        let column_idents = self
+            .columns
+            .iter()
+            .map(|(source_col, _target_col)| escape_identifier(source_col))
+            .collect::<Vec<String>>()
+            .join(", ");
+
+        let sql =
+        format!("SELECT {primary_key_idents}, {column_idents} FROM {quoted_table} ORDER BY {primary_key_idents}");
```
yujiyokoo, line 118: Is this safe? I'd prefer to use params over this

packages/cipherstash-proxy/src/cli/migrate/mod.rs
```diff
@@ -0,0 +1,309 @@
...
+                        );
+
+                        updated_count += 1;
+                        update_sql.push_str(&sql);
+                        records.extend(primary_key_vals.to_owned());
+                    }
+                    tokio_postgres::SimpleQueryMessage::CommandComplete(..) => {}
+                    tokio_postgres::SimpleQueryMessage::RowDescription(..) => {}
+                    _ => {
+                        error!(target: MIGRATE, msg = "Invalid row. You should not even be here.");
```
yujiyokoo, line 201:
```suggestion
                        error!(target: MIGRATE, msg = "Invalid row. This should be impossible.");
```

packages/cipherstash-proxy/src/cli/migrate/mod.rs
```diff
@@ -0,0 +1,309 @@
...
+        format!("SELECT {primary_key_idents}, {column_idents} FROM {quoted_table} ORDER BY {primary_key_idents}");
+
+        if self.commit() {
+            client.simple_query("BEGIN;").await?;
+        }
+
+        loop {
+            let offset = batch_count * self.batch_size;
+
+            let sql = format!("{sql} LIMIT {} OFFSET {offset} FOR UPDATE", self.batch_size);
```
yujiyokoo, line 127: Same as above

yujiyokoo (commented):
Looks good. Just one (or 2) question about the query building

packages/cipherstash-proxy/src/cli/migrate/mod.rs
tobyhede: I did have `#[cfg(test)]` but the module wasn't available in the integration tests.
I think because it is a dependency, so the integration suite is running with `cfg=test` but the dependency isn't?
Open to other ideas. Maybe I should import and wrap in the test suite instead

packages/cipherstash-proxy/src/cli/migrate/mod.rs
tobyhede: Is a valid point. 
Will raise as a separate issue, as this is all existing code.
Everything does get quoted appropriately, using the same quoting from the rust-postgres tools that are used to escape params.

packages/cipherstash-proxy/src/cli/migrate/mod.rs
yujiyokoo: Oh I see. That makes sense. Ta. Let's leave it this way.

packages/cipherstash-proxy/src/cli/migrate/mod.rs
yujiyokoo: Cool. I'm happy with addressing this in a separate issue.

yujiyokoo (approved):

PR #161, opened 2025-03-12
feat: Have the AWS RDS certificate bundle be optionally loaded into the system cert store.

damncabbage:
AWS RDS instances use a separate SSL CA and set of certificates to the Amazon root CA:
https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/UsingWithRDS.SSL.html

To connect via SSL to RDS, without changing the CA for the RDS instance, we need to pull in the AWS-provided certificate bundle.

To accomplish this, we have an entrypoint wrapper script (`cipherstash-proxy.sh`) that, at container start, optionally installs the bundle and then goes on to run the `cipherstash-proxy` binary. Set the `CS_DATABASE__INSTALL_AWS_RDS_CERT_BUNDLE` env var (eg. to `"true"`) for the container to load that bundle.

(The certificate bundle itself is fetched *at build time* for later run-time use; as long as the proxy image isn't extremely old, it should be fine.)

This PR builds on #159.

### DIFF
--- a/docker-entrypoint.sh
+++ b/docker-entrypoint.sh
@@ -1,0 +1,33 @@
+#!/usr/bin/env bash
+set -eu
+
+: "${CS_DATABASE__AWS_BUNDLE_PATH:=./aws-rds-global-bundle.pem}"
+
+# Optionally pull in the AWS RDS global certificate bundle. This is required
+# to communicate with AWS RDS instances, if they are not configured to use some
+# other certificates.
+# (This assumes that ca-certificates is installed, for csplit and update-ca-certificates.)
+case "${CS_DATABASE__INSTALL_AWS_RDS_CERT_BUNDLE:-}" in
+  # Have a guess at some common yaml-et-al encoding failures:
+  "") ;&
+  "false") ;&
+  "no") ;&
+  "0")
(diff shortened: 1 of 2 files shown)


docker-entrypoint.sh
```diff
@@ -0,0 +1,33 @@
...
+      exit 1
+    fi
+
+    echo "Installing AWS RDS certificate bundle..."
+    csplit --quiet --elide-empty-files --prefix /usr/local/share/ca-certificates/aws --suffix '.%d.crt' "$CS_DATABASE__AWS_BUNDLE_PATH" '/-----BEGIN CERTIFICATE-----/' '{*}'
+    update-ca-certificates
+    ;;
+esac
+
+exec cipherstash-proxy "$@"
```
damncabbage, line 33: Picked up this omission (lack of `"$@"`) from #157; cheers @auxesis.

tobyhede (approved):

auxesis (approved):
Nice work, thanks @damncabbage. ❤️

damncabbage:
Rebased off main again, switched the cert echo'ing to use stderr, tested the change in infradev; merging.

PR #162, opened 2025-03-12
build(mise): DEBUG flag now has the `build:docker` task retain its build output.

damncabbage:
Has the `mise run build:docker` task print its Docker build output (with `--progress=plain`), so it's possible to see said output after the run has finished.


coderdan (approved):

auxesis (approved):
That's neat, nice one @damncabbage.

PR #163, opened 2025-03-12
fix(eql): EQL identifier seralized with short field names

tobyhede:
Fix for  identifier "table" and "column"  properties to be serialized  into EQL JSON instead of the short names, "t" and "c"

### DIFF
--- a/packages/cipherstash-proxy/src/eql/mod.rs
+++ b/packages/cipherstash-proxy/src/eql/mod.rs
@@ -5,7 +5,7 @@ use cipherstash_client::{
 use serde::{Deserialize, Serialize, Serializer};
 use sqlparser::ast::Ident;
 
-#[derive(Clone, Debug, Deserialize, Serialize)]
+#[derive(Clone, Debug, Deserialize, PartialEq, Serialize)]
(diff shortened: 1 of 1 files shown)


packages/cipherstash-proxy/src/eql/mod.rs
```diff
@@ -106,3 +108,28 @@ where
...
+    pub fn plaintext_json() {
+        let identifier = Identifier::new("table", "column");
+        let pt = Plaintext {
+            identifier,
+            plaintext: "plaintext".to_string(),
+            version: 1,
+            for_query: None,
+        };
+
+        let value = serde_json::to_value(&pt).unwrap();
```
CDThomas, line 126: Do we serialize the `Plaintext` struct outside of tests? Might be more useful to test against the encrypted payload since that's what gets stored in PG.

CDThomas (approved):
Left one minor comment about testing, but LGTM otherwise.

packages/cipherstash-proxy/src/eql/mod.rs
tobyhede: Good call

PR #166, opened 2025-03-14
Refactor scope of config module

tobyhede:
[ x ]  extract schema and encrypt config logic into encrypt module.
[ x ]  extract config components into modules


packages/cipherstash-proxy/src/config/database.rs
```diff
@@ -0,0 +1,91 @@
...
+}
+
+impl DatabaseConfig {
+    pub fn default_host() -> String {
+        "127.0.0.1".to_string()
+    }
+
+    pub fn default_port() -> u16 {
+        5432
+    }
```
coderdan, line 40: Minor thing but you could make these as `const`.

```rust
pub const fn default_port() -> u16 {
```

packages/cipherstash-proxy/src/config/database.rs
```diff
@@ -0,0 +1,91 @@
...
+    pub config_reload_interval: u64,
+
+    #[serde(default = "DatabaseConfig::default_schema_reload_interval")]
+    pub schema_reload_interval: u64,
+}
+
+impl DatabaseConfig {
+    pub fn default_host() -> String {
+        "127.0.0.1".to_string()
+    }
```
coderdan, line 36: Could this return a `&'static str`?

packages/cipherstash-proxy/src/config/database.rs
```diff
@@ -0,0 +1,91 @@
...
+            }
+        })?;
+        Ok(name)
+    }
+}
+
+///
+/// Password is NEVER EVER displayed
+///
+impl Display for DatabaseConfig {
```
coderdan, line 83: I recommend the use of [opaque-debug](https://crates.io/crates/opaque-debug).
Or could be a good opportunity to use [vitaminc-protected](https://crates.io/crates/vitaminc-protected).

packages/cipherstash-proxy/src/config/log.rs
```diff
@@ -0,0 +1,211 @@
...
+
+    #[serde(default = "LogConfig::default_log_level")]
+    pub schema_level: LogLevel,
+}
+
+#[derive(Clone, Copy, Debug, Deserialize, PartialEq, ValueEnum)]
+#[serde(rename_all = "lowercase")]
+pub enum LogFormat {
+    // Serde does not seem to have a case insensitive option. alias is clunky, but better than custom de/serialisers
+    #[serde(alias = "Pretty", alias = "pretty", alias = "PRETTY")]
```
coderdan, line 61: TIL about alias.

packages/cipherstash-proxy/src/config/server.rs
```diff
@@ -0,0 +1,84 @@
...
+        "0.0.0.0".to_string()
+    }
+
+    pub fn default_port() -> u16 {
+        DEFAULT_PORT
+    }
+
+    pub fn default_shutdown_timeout() -> u64 {
+        DEFAULT_SHUTDOWN_TIMEOUT
+    }
```
coderdan, line 52: Same comments as before.

packages/cipherstash-proxy/src/config/tls.rs
```diff
@@ -0,0 +1,195 @@
...
+///
+/// Server TLS Configuration
+/// This is listener/inbound connection config
+///
+#[derive(Clone, Debug, Deserialize)]
+#[serde(tag = "type")]
+pub enum TlsConfig {
+    Pem {
+        certificate: String,
+        private_key: String,
```
coderdan, line 18: Not now necessarily but this would be a good candidate for VitaminC Protected as well.
Either that or we perhaps consider using Zeroize directly.

coderdan (approved):
A few nits and a couple of things to consider for future work.
Otherwise LGTM!

packages/cipherstash-proxy/src/config/log.rs
tobyhede: @coderdan To be fair, it's a pretty ugly workaround. Alternative was writing it from scratch, but just seemed not worth the effort to handle typos in log config options.

packages/cipherstash-proxy/src/config/database.rs
tobyhede: Has to be turned into a String anyway to be set on the struct field.

packages/cipherstash-proxy/src/config/database.rs
tobyhede: Pushed change using vitaminc for database password.

PR #168, opened 2025-03-17
docs: explicitly state that docker daemon should be started

freshtonic:
Explicitly state in `DEVELOPMENT.md` that the Docker daemon should be launched.

## Acknowledgment

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -3,7 +3,7 @@
 > [!IMPORTANT]
 > **Before you quickstart** you need to have this software installed:
 >  - [mise](https://mise.jdx.dev/) — see the [installing mise](#installing-mise) instructions
->  - [Docker](https://www.docker.com/) — see Docker's [documentation for installing](https://docs.docker.com/get-started/get-docker/)
+>  - [Docker](https://www.docker.com/) — see Docker's [documentation for installing](https://docs.docker.com/get-started/get-docker/) and ensure that your Docker daemon is running. On *macOS* and *Linux* you can do this by running Docker Desktop.  See: [Docker Desktop docs](https://docs.docker.com/get-started/introduction/get-docker-desktop/). If you have installed Docker without Docker Desktop then you will need to launch `dockerd` manually.
 
 Local development quickstart:
 


DEVELOPMENT.md
```diff
@@ -19,6 +19,18 @@ sh preflight.sh
 mise trust --yes
 mise trust --yes tests
 mise install
+```
```
CDThomas, line 22: This reads to me now like only the first code block is needed for the "Local development quickstart" section (when both code blocks are still necessary for getting started).

Might be clearer to hoist the new info into the "Before you quickstart" callout or to make "Local development quickstart" a proper heading.

freshtonic: Ah yep, makes sense!

DEVELOPMENT.md
```diff
@@ -3,7 +3,7 @@
 > [!IMPORTANT]
 > **Before you quickstart** you need to have this software installed:
 >  - [mise](https://mise.jdx.dev/) — see the [installing mise](#installing-mise) instructions
->  - [Docker](https://www.docker.com/) — see Docker's [documentation for installing](https://docs.docker.com/get-started/get-docker/)
+>  - [Docker](https://www.docker.com/) — see Docker's [documentation for installing](https://docs.docker.com/get-started/get-docker/) and ensure that your Docker daemon is running. On *macOS* and *Linux* you can do this by running Docker Desktop.  See: [Docker Desktop docs](https://docs.docker.com/get-started/introduction/get-docker-desktop/). If you have installed Docker without Docker Desktop then you will need to launch `dockerd` manually.
```
CDThomas, line 6: Awesome, only other minor suggestion is to split sentences across lines for consistency with the formatting that we use elsewhere in markdown files. Something like:
```suggestion
>  - [Docker](https://www.docker.com/) — see Docker's [documentation for installing](https://docs.docker.com/get-started/get-docker/) and ensure that your Docker daemon is running.
>    On *macOS* and *Linux* you can do this by running Docker Desktop. 
>    See: [Docker Desktop docs](https://docs.docker.com/get-started/introduction/get-docker-desktop/).
>    If you have installed Docker without Docker Desktop then you will need to launch `dockerd` manually.
```

CDThomas (approved):
Left one more minor suggestion around formatting but the content LGTM 👍

DEVELOPMENT.md
auxesis: Annoyingly, newline behaviour is different in [alert blocks](https://github.blog/changelog/2023-12-14-new-markdown-extension-alerts-provide-distinctive-styling-for-significant-content/) than the rest of markdown. 

@CDThomas's proposed changes look like this: 

image

For now, best to keep the sentences joined. 🤦

CDThomas: Silly me for assuming that newlines in alert blocks would behave like regular markdown quotes 🤦 Good catch @auxesis.

PR #172, opened 2025-03-17
chore: remove old config

tobyhede:
Config now uses the same naming as `stash setup` 

This removes the old references.


yujiyokoo (approved):
LGTM :)

PR #174, opened 2025-03-18
feat: improve error message for internal errors

yujiyokoo:
This changes error mapping target of EqlMapper::InternalError from `MappingError::StatementCouldNotBeMapped` to a newly created error type `MappingError::Internal`. Also adds a new section in the docs.

I think this is better as "statement could not be mapped" can be user-side issue (like 4XX errors in HTTP), but internal error is server-side (like 5XX in HTTP).

It still does the same logging and counting as `StatementCouldNotBeMapped` but the raised (mapped) error and its messages are distinct from `StatementCouldNotBeMapped`.

## Acknowledgment

### DIFF
--- a/docs/errors.md
+++ b/docs/errors.md
@@ -170,13 +170,13 @@ Check the supported types for encrypted columns.
 
 ## Internal Mapper error <a id='mapping-internal-error'></a>
 
-An error that is internal to the EQL Mapper module. This may be a defect or a limitation of the current EQL Mapper.
-
+An internal error occurred when attempting to rewrite the SQL statement.
+This could be due to an internal invariant failure or because of a specific fragment of unsupported SQL syntax.
 
 ### Error message
 
 ```
-Internal error: 'internal details'
+Statement encountered an internal error. This may be a bug in the statement mapping module of CipherStash Proxy.
 ```
 
 ### How to Fix
--- a/docs/errors.md
+++ b/docs/errors.md
@@ -12,6 +12,7 @@
   - [Invalid parameter](#mapping-invalid-parameter)
   - [Invalid SQL statement](#mapping-invalid-sql-statement)
   - [Unsupported parameter type](#mapping-unsupported-parameter-type)
+  - [Internal Error](#mapping-internal-error)
 
 - Encrypt Errors:
   - [Column could not be encrypted](#encrypt-column-could-not-be-encrypted)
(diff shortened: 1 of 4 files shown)


packages/cipherstash-proxy/src/error.rs
```diff
@@ -6,7 +6,7 @@ use std::io;
 use thiserror::Error;
 use tokio::time::error::Elapsed;
 
-const ERROR_DOC_BASE_URL: &str = "https://github.com/cipherstash/proxy/docs/errors.md";
+const ERROR_DOC_BASE_URL: &str = "https://github.com/cipherstash/proxy/blob/main/docs/errors.md";
```
yujiyokoo, line 9: Without `blob/main`, it's not found

tobyhede: Oh. That's a bit annoying.

packages/cipherstash-proxy/src/postgresql/frontend.rs
```diff
@@ -737,6 +737,16 @@ where
 
                 Ok(typed_statement)
             }
+            Err(EqlMapperError::InternalError(str)) => {
+                warn!(
```
tobyhede, line 741: Might be able to drop this `warn` as I think the error should be logged as error higher in the stack?

docs/errors.md
```diff
@@ -164,6 +165,28 @@ Check the supported types for encrypted columns.
 
 
 
+<!-- ---------------------------------------------------------------------------------------------------- -->
+
+
+## Internal Mapper error <a id='mapping-internal-error'></a>
+
+An error that is internal to the EQL Mapper module. This may be a defect or a limitation of the current EQL Mapper.
```
tobyhede, line 173: Not sure EQL Mapper is something we mention anywhere, but not sure what the customer-facing language should be. @kateandrews @auxesis any thoughts?

freshtonic: @yujiyokoo @tobyhede maybe something like this?

"An internal error occurred when attempting to rewrite the SQL statement. This could be due to an internal invariant failure or because of a specific fragment of unsupported SQL syntax"

packages/cipherstash-proxy/src/error.rs
freshtonic: Ideally, we should ascertain the current version of the proxy and link to a specific commit so that the release and the docs we link to are in sync.

However, that would require some proc_macro shenanigans to interpolate the appropriate tag into the URL so I would be happy to shave that yak later.

docs/errors.md
yujiyokoo: @freshtonic that sounds good to me. I'll use that unless there's any other suggestion from other people

packages/cipherstash-proxy/src/postgresql/frontend.rs
yujiyokoo: Leaving it here for now as Proxy currently won't report mapping errors unless `enable_mapping_errors` is set

PR #180, opened 2025-03-19
fix: make connection timeout optional

tobyhede:
Replicate postgres behaviour by making connection timeout optional. 

Timeout is still configurable, but default is no timeout.


auxesis (approved):
Thanks for this one @tobyhede!

PR #181, opened 2025-03-19
feat(proxy): output installed EQL version on startup

freshtonic:
This queries the database with `select cs_eql_version();` (rather than the version at build time) for accuracy.

## Acknowledgment

### DIFF
--- a/mise.local.example.toml
+++ b/mise.local.example.toml
@@ -14,6 +14,8 @@ CS_DEFAULT_KEYSET_ID = "default-keyset-id"
 CS_CLIENT_KEY = "client-key"
 CS_CLIENT_ID = "client-id"
 
+# The release of EQL that the proxy tests will use and releases will be built with
+CS_EQL_VERSION="eql-0.5.4"
 
 # TLS variables are required for providing TLS to Proxy's clients.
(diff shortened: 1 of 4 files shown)


auxesis (approved):
Looks good, thanks @freshtonic. 

I can't see where the version string is output, but the real value is in #182 anyway. 💪

freshtonic:
> Looks good, thanks @freshtonic.
> 
> I can't see where the version string is output, but the real value is in #182 anyway. 💪

It's output in the `info!` message here: https://github.com/cipherstash/proxy/pull/181/files#diff-5eb487b6c73a46fb09652d308e672bc07ca1aaf44986b2352597b55d71d63790L214-L220

auxesis:
> > Looks good, thanks @freshtonic.
> > I can't see where the version string is output, but the real value is in #182 anyway. 💪
> 
> It's output in the `info!` message here: https://github.com/cipherstash/proxy/pull/181/files#diff-5eb487b6c73a46fb09652d308e672bc07ca1aaf44986b2352597b55d71d63790L214-L220

Cool, right in front of me the whole time! 🤦‍♂️

Thanks @freshtonic!

PR #183, opened 2025-03-19
fix: ensure pem data is never logged

tobyhede:
Refactored TlsConfig to return `Error` if invalid on the startup check. 

We can use `thiserror` handles the error string generation, and the error is more easily kept consistent.

I've also differentiated the config options for path and pem in the field names. 

```
[tls]
# Certificate path
# Env: CS_TLS__CERTIFICATE_PATH
certificate_path = "./server.cert"

# Private Key path
# Env: CS_TLS__PRIVATE_KEY_PATH
private_key_path = "./server.key"

# Certificate path
# Env: CS_TLS__CERTIFICATE_PEM
certificate_pem = "..."

# Private Key path
# Env: CS_TLS__PRIVATE_KEY_PEM
private_key_pem = "..."
```

I think having the name be explicit might reduce the potential for footguns.


freshtonic (approved):
LGTM

PR #189, opened 2025-03-22
docs: improve config error messaging

tobyhede:
Reworks the config error message handling to provide more granular error messages containing the specific field and keys that are required, and link to the config section of the README.


packages/cipherstash-proxy/src/config/tandem.rs
```diff
@@ -254,18 +259,31 @@ impl Default for PrometheusConfig {
     }
 }
 
+// static RE: LazyLock<Regex> =
+//     LazyLock::new(|| Regex::new(r"missing field `([^`]+)`(.+)`([^`]+)`").unwrap());
```
yujiyokoo, line 263: Is this just an older version hanging around?

yujiyokoo: I'm also wondering if it's better to check for "missing" (or "missing field") if the code assumes it's a missing field. This may be less confusing if anybody adds "invalid field" or something in the future, but might not be helpful either, so I'm not sure.

yujiyokoo (approved):
Nice work!

packages/cipherstash-proxy/src/config/tandem.rs
tobyhede: I actually pulled the config crate to see if I could add more structure to the errors, but it seemed like it would require some pretty invasive changes.
Added extra tests instead. If the underlying errors change, we should catch.

PR #193, opened 2025-03-25
Warn if result column is of type cs encrypted v1

yujiyokoo:
This warns if there is what looks like an encrypted column but no encrypt config for the column.

It differenciates between implicit passthrough (encrypt config missing) and explicitly set passthrough (mapping disabled) and does not warn if explicitly set t passthrough.

Originally the idea was to warn the user when the column type is the domain type `cs_encrypted_v1` but as far as I can tell, in `RowDescription` messages, we see the type oid for JSONB, not the domain.

So instead, we attempt to parse the column value as `cs_encrypted_v1` and warn the user if that is successful but with no encrypt config.

## Acknowledgment

### DIFF
--- a/packages/cipherstash-proxy/src/postgresql/backend.rs
+++ b/packages/cipherstash-proxy/src/postgresql/backend.rs
@@ -8,7 +8,7 @@ use crate::connect::Sender;
 use crate::encrypt::Encrypt;
 use crate::eql::Encrypted;
 use crate::error::Error;
-use crate::log::{DEVELOPMENT, MAPPER, PROTOCOL};
+use crate::log::{CONFIG, DEVELOPMENT, MAPPER, PROTOCOL};
 use crate::postgresql::context::Portal;
 use crate::postgresql::messages::data_row::DataRow;
 use crate::postgresql::messages::param_description::ParamDescription;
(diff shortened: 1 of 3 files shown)


packages/cipherstash-proxy/src/postgresql/backend.rs
```diff
@@ -91,6 +91,8 @@ where
         let sent: u64 = bytes.len() as u64;
         counter!(SERVER_BYTES_RECEIVED_TOTAL).increment(sent);
 
+        self.warn_if_encrypted_value_with_no_config(&code, &bytes)?;
```
tobyhede, line 94: Handling the check here means double-handling for encrypted statements as we iterate and parse the data again. 

An encrypted statement will only contain an unencrypted column if the configuration for one column was removed `after` a record was encrypted. 

I think the most likely situation for customers is that they accidentally remove the config for the whole table, rather than a single column. In this case, we could display the warning just for `passthrough` statements.

tobyhede:
Possible alternative ... `RowDescription` has some fields that might enable us to check the schema. 

```
    pub name: String,
    pub table_oid: i32,
    pub table_column: i16,
    pub type_oid: postgres_types::Type,
```

I think name might be any `alias` in the sql, not the actual column name, so not totally useful. 
But the `table_oid` and `table_column` could be mapped to the schema.

yujiyokoo:
Done in #196 instead. Closing this.

PR #196, opened 2025-03-27
Error or warn if encrypt config missing

yujiyokoo:
This changes the missing config logic to:

```
IF mapping is disabled
  pass through the query
ELSE IF column is to be encrypted but no config for it is found
  IF mapping_errors_enabled
    return error
  ELSE
    output warning and pass through the query
  END
END
```

It also adds a single test for the warning behaviour. There is no test for the error as those integration tests increase the CI time (I have tested manually though).

## Acknowledgment


packages/cipherstash-proxy-integration/src/extended_protocol_error_messages.rs
```diff
@@ -66,7 +66,11 @@ mod tests {
         if let Err(err) = result {
             let msg = err.to_string();
 
-            assert_eq!(msg, "db error: ERROR: Column 'encrypted_unconfigured' in table 'unconfigured' has no Encrypt configuration. For help visit https://github.com/cipherstash/proxy/blob/main/docs/errors.md#encrypt-unknown-column");
+            // This is similar to below. The error message comes from tokio-postgres when Proxy
+            // returns cs_encrypted_v1 and the client cannot convert to a string.
+            // If mapping errors are enabled (enable_mapping_errors or CS_DEVELOPMENT__ENABLE_MAPPING_ERRORS),
+            // then Proxy will return an error that says "Column X in table Y has no Encrypt configuration"
+            assert_eq!(msg, "error serializing parameter 1: cannot convert between the Rust type `&str` and the Postgres type `cs_encrypted_v1`");
```
yujiyokoo, line 73: Unfortunately, this is not as clear now, but with mapping errors disabled (default, and how it is here) we emit a warning, and return the value to the client in `cs_encrypted_v1`, and this is the client (tokio-postgres) reporting an error. Not sure what else we can do to improve

tests/python/tests/test_error_messages.py
```diff
@@ -54,7 +54,10 @@ def test_encrypted_column_with_no_configuration():
 
                 sql = "INSERT INTO unconfigured (id, encrypted_unconfigured) VALUES (%s, %s)"
 
-                with pytest.raises(psycopg.Error, match='#encrypt-unknown-column'):
+                # This is EQL catching the error and returning it. Details are in docs/errors.md
+                # When mapping errors are enabled, (enable_mapping_errors or CS_DEVELOPMENT__ENABLE_MAPPING_ERRORS)
+                # Proxy will return an error that says "Column X in table Y has no Encrypt configuration"
+                with pytest.raises(psycopg.Error, match=r"Encrypted column missing \w+ \(\w+\) field"):
```
yujiyokoo, line 60: Similarly to the other test above, we show EQL's error here when trying to insert, as we emit the warning and let the unencrypted record pass through to be caught by Postgres.

packages/cipherstash-proxy/src/postgresql/frontend.rs
```diff
@@ -86,7 +86,7 @@ where
         let sent: u64 = bytes.len() as u64;
         counter!(CLIENTS_BYTES_RECEIVED_TOTAL).increment(sent);
 
-        if self.encrypt.is_passthrough() {
+        if self.encrypt.config.mapping_disabled() {
```
yujiyokoo, line 89: `is_passthrough` checks if mapping is disabled or encrypt config is empty. Here it's updated to only one

packages/cipherstash-proxy-integration/src/extended_protocol_error_messages.rs
tobyhede: I've run into this a couple of times with tokio-postgres and it can be a bit confusing. 
Not much we can do, because is actually tokio enforcing type correctness - the statement types have not been rewritten.

tobyhede (approved):

PR #209, opened 2025-04-22
feat: SQL transformations related to `GROUP BY`

freshtonic:
Includes a version bump to the latest `sqltk` version which vendors

`sqlparser` as `sqltk-parser` and includes an extra parameter in
`Transform::transform` to allow contextual transformations (instead of
solely by type).

The following refactorings were done as part of this PR:

1. Break sole `Transform` impl into modular rules using the
   `TransformationRule` trait.

2. Removed `EqlFunctionTracker` - this existed to support `Expr`
   transformations in `ORDER BY` clauses. With contextual transformation
   it was no longer required.

3. Removed `TypeCell` - it did not carry its weight and made debugging
   really hard. The `TypeRegistry` now associates nodes to Arc<Type>
   using an intermediary `TypeVar`.

An extra method was added to the post type checking step in the
`EqlMapper` to resolve all `Value` nodes which have no concrete type
after the inference step to `NativeValue`.

Additionally, I went through many levels of Hell trying to figure out how I 
broke some existing tests so this PR also includes:

feat: tracing of `InferType` impls & `Unifier::unify`

Debugging type checking code in VSCode is suboptimal - we really want a
graphical representation of inference at every step (but also a debugger
that doesn't lock up at inconvenient times would be great).

To make instrumentation consistent I implemented a `#[trace_infer]`
attribute macro which can be used on `InferType` impls.

The `Unifier` has a single `#[tracing::instrument!(..)]` call and there
is also code in `EqlMapper` to dump nodes types and type substitutions
at the end of the type checking process which is useful when debugging
failing inference tests.

To get a trace from a test, uncomment the `init_tracing();` line.

The output is *very* verbose but very useful.

## Acknowledgment


tobyhede (approved):

tobyhede:
Is it possible to make the tracing optional? Maybe use a crate feature?
I assume it will add to compile time, will it add much overhead to release build?

freshtonic:
> Is it possible to make the tracing optional? Maybe use a crate feature? I assume it will add to compile time, will it add much overhead to release build?

Disabling tracing of inference at compile time is definitely possible.

It doesn't seem to significantly affect compilation times FWIW. As for runtime overhead, my understanding is that it's practically a no-op when there's no subscriber attached.

tobyhede (approved):

PR #212, opened 2025-04-30
feat: Upgrades to sqltk v0.6.0 

damncabbage:
This PR:
- Upgrades to sqltk v0.6.0, which includes sqltk-parser exported as `sqltk::parser`.
- Makes the corresponding adjustments to accommodate the parser `0.52.0` -> `0.53.0-...` upgrade.
- Removes the direct dependency on `sqltk-parser`, using the re-export instead.


freshtonic (approved):
Yep, looks fairly mechanical.

The token span shenanigans add a bit of noise but I can't see how we can avoid that.

LGTM! Thanks @damncabbage

PR #215, opened 2025-04-30
Docs: Add link to request x86_64 Docker image

kateandrews:
## Acknowledgment

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,6 +54,10 @@ Behind the scenes, it uses the [Encrypt Query Language](https://github.com/ciphe
 > [!IMPORTANT]
 > **Prerequisites:** Before you start you need to have this software installed:
 >  - [Docker](https://www.docker.com/) — see Docker's [documentation for installing](https://docs.docker.com/get-started/get-docker/)
+> 
+> **Supported architectures:** We currently support `arm64v8` architectures. If you're interested in a `x86_64` Docker image, upvote [this idea](https://github.com/cipherstash/proxy/discussions/214).
+
+ 
 
 Get up and running in local dev in < 5 minutes:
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -54,6 +54,10 @@ Behind the scenes, it uses the [Encrypt Query Language](https://github.com/ciphe
 > [!IMPORTANT]
 > **Prerequisites:** Before you start you need to have this software installed:
 >  - [Docker](https://www.docker.com/) — see Docker's [documentation for installing](https://docs.docker.com/get-started/get-docker/)
+> 
+> **Supported architectures:** We currently support `arm64v8` architectures. If you're interested in a `x86_64` Docker image, upvote [this idea](https://github.com/cipherstash/proxy/discussions/214).
```
auxesis, line 58:
```suggestion
> **Supported architectures:** CipherStash Proxy provides Docker images for these architectures: `linux/arm64`
> If you're interested in a Docker image for other architectures (like `linux/amd64`), upvote [this idea](https://github.com/cipherstash/proxy/discussions/214).
```

auxesis: Also, I think this might be better in a [reference doc](https://github.com/cipherstash/proxy/blob/main/docs/reference.md), rather than in the main README. 

It's too low a level of detail for a getting started guide, but it's the sort of thing that makes sense in a reference doc.

auxesis (changes requested):
This is a good start, and we should move it to a reference doc.

docs/reference.md
```diff
@@ -266,6 +267,14 @@ If the proxy is running on a host other than localhost, access on that host.
 | `cipherstash_proxy_statements_total`                  | Counter   | Total number of SQL statements processed by CipherStash Proxy               |
 | `cipherstash_proxy_statements_unmappable_total`       | Counter   | Total number of unmappable SQL statements processed by CipherStash Proxy    |
 
+## Supported architectures
+
+CipherStash Proxy provides Docker images for `linux/arm64` architectures.
```
auxesis, line 272:
```suggestion
CipherStash Proxy is [available as a Docker container image](https://hub.docker.com/r/cipherstash/proxy) for `linux/arm64` architectures.
```

auxesis (approved):
Thanks @kateandrews! 

I have added one small suggested copy change.

PR #217, opened 2025-05-07
Expand passthrough test cases

tobyhede:
Adds additional passthrough tests to cover some of the more obscure SQL syntax encountered in the wild. 

Covers
- insert from select
- insert with value from select
- insert with returning
- delete with select subquery


auxesis (approved):
😍 thanks @tobyhede!

PR #224, opened 2025-05-12
Eql v2 decryption

tobyhede:
EQL 2.0 changes the internal type representation of encrypted records to use a PostgreSQL composite type.
Returned values are wrapped as a tuple `("{ ... }")`
When decrypting returned records, the json needs to be extracted.


freshtonic (approved):
Hairy. LGTM.

PR #229, opened 2025-05-13
Cast literals and params to `eql_v1_encrypted`

tobyhede:
Fix for SQL error when using `ore` or `unique` index terms:
```
  operator is not unique: eql_v1_encrypted > record
```

Encrypted parameters and literals are currently being wrapped into a `ROW` which create a value of type `record`.

This explicitly casts literals and params as `::jsonb::eql_v1_encrypted`


freshtonic (approved):
Nice!

PR #230, opened 2025-05-15
Timestamp data type

tobyhede:
- adds handling for timestamp data type from_sql 
- test to map timestamp 
- add missing biguint to_sql implementation at no extra charge


packages/cipherstash-proxy-integration/src/map_params.rs
```diff
@@ -182,6 +182,35 @@ mod tests {
...
     }
 
+    #[tokio::test]
+    async fn map_timestamp() {
+        trace();
+
+        let client = connect_with_tls(PROXY).await;
+
+        let id = id();
+        let encrypted_timestamp = "2000-10-01T17:30:00Z".parse::<DateTime<Utc>>().unwrap();
```
yujiyokoo, line 192: Would it be a good idea to add another test case set in a different time zone than UTC?

packages/cipherstash-proxy/src/postgresql/data/to_sql.rs
```diff
@@ -47,8 +49,10 @@ fn binary_to_sql(plaintext: &Plaintext) -> Result<BytesMut, Error> {
         Plaintext::Utf8Str(x) => x.to_sql_checked(&Type::TEXT, &mut bytes),
         Plaintext::JsonB(x) => x.to_sql_checked(&Type::JSONB, &mut bytes),
         Plaintext::Decimal(x) => x.to_sql_checked(&Type::NUMERIC, &mut bytes),
-        // TODO: Implement these
-        Plaintext::BigUInt(_x) => unimplemented!(),
+        Plaintext::BigUInt(x) => {
```
yujiyokoo, line 52: Was this needed here for timestamp, or is this a separate implementation? If so, is this covered with a test case?

yujiyokoo (approved):
LGTM

tobyhede:
Closing and will reopen.

PR #231, opened 2025-05-15
feat: check column configuration against encrypted column before decrypt

tobyhede:
Check the identifier in the returned ciphertext matches the identifier in the column configuration before decryption. 
Returns an error if a mismatch is detected. 
Helps protect against confused deputy attacks.

### DIFF
--- a/packages/cipherstash-proxy/src/postgresql/backend.rs
+++ b/packages/cipherstash-proxy/src/postgresql/backend.rs
@@ -4,11 +4,12 @@ use super::message_buffer::MessageBuffer;
 use super::messages::error_response::ErrorResponse;
 use super::messages::row_description::RowDescription;
 use super::messages::BackendCode;
+use super::Column;
 use crate::connect::Sender;
 use crate::encrypt::Encrypt;
 use crate::eql::EqlEncrypted;
-use crate::error::Error;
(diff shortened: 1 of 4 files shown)


packages/cipherstash-proxy/src/postgresql/backend.rs
```diff
@@ -313,6 +316,41 @@ where
         Ok(())
     }
 
+    fn check_column_config(
+        &mut self,
+        projection_columns: &Vec<Option<Column>>,
+        ciphertexts: &Vec<Option<EqlEncrypted>>,
```
coderdan, line 322: Can this be a slice?

packages/cipherstash-proxy/src/postgresql/backend.rs
```diff
@@ -313,6 +316,41 @@ where
...
+    ) -> Result<(), Error> {
+        for (col, ct) in projection_columns.iter().zip(ciphertexts) {
+            // debug!(target: DECRYPT, client_id = self.context.client_id, ?col, ?ct);
```
coderdan, line 325: Can this be removed?

coderdan (approved):
Nice. Looks good to me.
The other part of the confused deputy problem is to include the identifier in the lock context (or in the aad of the local ciphertext at least). Not sure if that's been done/carded up but would be a necessary element.

PR #233, opened 2025-05-21
fixes for rust 1.87 clippy lints

tobyhede:


packages/eql-mapper/src/inference/type_error.rs
```diff
@@ -34,14 +31,15 @@ pub enum TypeError {
     #[error("{}", _0)]
     SchemaError(#[from] SchemaError),
 
-    #[error("Cannot unify node types for nodes:\n 1. node: {} type: {}\n 2. node: {} type: {}\n error: {}", _1, _2, _3, _4, _0)]
-    OnNodes(Box<TypeError>, String, Arc<Type>, String, Arc<Type>),
-
+    // #[error("Cannot unify node types for nodes:\n 1. node: {} type: {}\n 2. node: {} type: {}\n error: {}", _1, _2, _3, _4, _0)]
+    // OnNodes(Box<TypeError>, String, Arc<Type>, String, Arc<Type>),
```
yujiyokoo, line 35: Can we have this deleted, or maybe have a little info on why it's here?

yujiyokoo (approved):
LGTM. 1 question/comment about a comment

packages/eql-mapper/src/inference/type_error.rs
tobyhede: My bad, I was experimenting to make it work

PR #237, opened 2025-05-23
Handle special characters in db config values.

tobyhede:
Use tokio_postgres::Config instead of connection string.
This change turns proxy db config into tokio_postgres::Config.

Note: this changes the test database password.

Rather than introduce new configuration for testing special characters, this adds a special character to the default config so testing is baked in.

### DIFF
--- a/docker-entrypoint.sh
+++ b/docker-entrypoint.sh
@@ -1,14 +1,28 @@
 #!/usr/bin/env bash
 set -eu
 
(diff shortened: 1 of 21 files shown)


docker-entrypoint.sh
```diff
@@ -1,14 +1,28 @@
...
+      *)
+        printf '%%%02X' "'$c"
+        ;;
+    esac
+  done
+}
+
 wait_for_postgres_or_exit() {
-  host=${CS_DATABASE__HOST}
+  host=${CS_DATABASE__HOST}x
```
CDThomas, line 25: Looks like there's a typo here:
```suggestion
  host=${CS_DATABASE__HOST}
```

tobyhede: I pushed a fix for this.
Turns out I didn't commit the fix.

PR #239, opened 2025-05-26
Fix for "Error: insufficient data left in message" with null values

tobyhede:
Fixes issue with encoding `NULL` parameter values in Bind messages.

`NULL` params are of length 0, but the length is encoded in a Bind message as `-1`.


freshtonic (approved):
Another nasty bug slayed 💪

PR #240, opened 2025-05-26
OIDC

cisopaul:
Removing debug from the workflow action, but leaving the repo/context/user output only

## Acknowledgment


auxesis (approved):
LGTM, thanks @cisopaul!

PR #242, opened 2025-05-27
fix: url encode username in docker entrypoint

tobyhede:


auxesis (approved):
Nice catch, thanks @tobyhede.

PR #244, opened 2025-05-27
docs: Adds TLS and Docker configuration docs.

damncabbage:
Adds documentation covering:
- The `[tls]` section of Proxy config,
- Clarification of `require_tls` and `with_tls_verification`, and
- The environment variables the Proxy Docker entrypoint and dev `docker-compose.yml` uses.

## Acknowledgment

### DIFF
--- a/docs/reference.md
+++ b/docs/reference.md
@@ -229,7 +229,7 @@ CS_DATABASE__SCHEMA_RELOAD_INTERVAL="10"
 
 ### Docker-specific configuration
 
-As a convenience for development, if you use [Proxy's Docker container](../proxy.Dockerfile) with its default entrypoint and the below environment variables set, the EQL SQL will be applied to the database, and an example schema (for example, with the `users` table, from the [README Getting Started example](../README.md#getting-started)) will be loaded. These are turned on by default in the [development `docker-compose.yml`](../docker-compose.yml):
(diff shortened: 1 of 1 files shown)


damncabbage:
oop, thank you!

docs/reference.md
```diff
@@ -207,17 +216,35 @@ Although Proxy attempts to detect the environment and set a sensible default for
...
+As a convenience for development, if you use [Proxy's Docker container](../proxy.Dockerfile) with its default entrypoint and the below environment variables set, the EQL SQL will be applied to the database, and an example schema (for example, with the `users` table, from the [README Getting Started example](../README.md#getting-started)) will be loaded. These are turned on by default in the [development `docker-compose.yml`](../docker-compose.yml):
+
+(It is not recommended to use either of these in production.)
+
+```bash
+CS_DATABASE__INSTALL_EQL="true"
+CS_DATABASE__INSTALL_EXAMPLE_SCHEMA="true"
+```
+
+As a convenience for production, if you use [Proxy's Docker container](../proxy.Dockerfile) with the default entrypoint and the below environment variable set, the AWS RDS global certificate bundle will be added to the operating system's set of trusted certificates. This is recommended when running Proxy on AWS.
```
yujiyokoo, line 241: The "As a convenience for production, if you use [Proxy's Docker container](../proxy.Dockerfile) with the default entrypoint and the below environment variable set" part feels repetitive. Maybe skipping that and say "The Proxy container will also set the AWS RDS..." here?

yujiyokoo (approved):
LGTM

docs/reference.md
damncabbage: The repetition was intentional, but it's had the opposite effect to what I intended. The first one is "As a convenience for **development**", the second one is "As a convenience for **production**". I'll distinguish them better, pardon.

yujiyokoo: Oops I didn't see that 😅

PR #245, opened 2025-05-27
Feature/prefer crn over workspace_id and region

yujiyokoo:
This makes workspace CRN the preferred way to configure proxy, though workspace ID and region are still supported.

* makes the CRN the default way to store region and workspace Id in AuthConfig
* adds custom deserialiser to handle more complex scenarios of supporting multiple configuration methods
* also splits the ZeroKMS config for ease of testing, and adds some tests
* adds some test toml files so both env vars and toml based configuration can be tested

Now ignores workspace_id and region but outputs warnings:

Screenshot 2025-05-29 at 4 47 20 PM

## Acknowledgment

### DIFF
--- a/packages/cipherstash-proxy/src/config/tandem.rs
+++ b/packages/cipherstash-proxy/src/config/tandem.rs
@@ -816,6 +816,28 @@ mod tests {
     }
 
     #[test]
+    fn crn_can_load_from_env_vars() {
+        let env = merge_env_vars(vec![(
+            "CS_WORKSPACE_CRN",
+            Some("crn:us-west-1.aws:E4UMRN47WJNSMAKR"),
+        )]);
+
+        with_no_cs_vars(|| {
+            temp_env::with_vars(env, || {
+                let config = TandemConfig::build("tests/config/unknown.toml").unwrap();
+                assert_eq!(
+                    "E4UMRN47WJNSMAKR",
+                    config.auth.workspace_crn.workspace_id.to_string()
+                );
+                assert_eq!(
+                    "us-west-1.aws",
+                    config.auth.workspace_crn.region.to_string()
+                );
+            })
+        })
+    }
+
+    #[test]
     fn crn_from_toml_with_workspace_id_and_region_env_vars() {
         let env = merge_env_vars(vec![
             ("CS_WORKSPACE_ID", Some("DCMBTGHEX5R2RMR4")),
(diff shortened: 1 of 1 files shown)


Cargo.lock
```diff
@@ -650,9 +650,9 @@ dependencies = [
 
 [[package]]
 name = "cipherstash-client"
-version = "0.22.0"
+version = "0.22.1"
```
tobyhede, line 653: Nit: I think we should be explict with the version in Cargo.toml

packages/cipherstash-proxy/src/test_helpers.rs
```diff
@@ -0,0 +1,77 @@
+use std::io;
```
tobyhede, line 1: Nice refactor 👍

packages/cipherstash-proxy/src/config/tandem.rs
```diff
@@ -35,10 +38,53 @@ pub struct TandemConfig {
...
 
+fn deserialise_auth_config<'de, D>(deserializer: D) -> Result<AuthConfig, D::Error>
+where
+    D: serde::Deserializer<'de>,
+{
+    #[derive(Deserialize)]
+    struct AuthConfigRaw {
+        workspace_crn: Option<String>,
+        workspace_id: Option<String>,
+        region: Option<String>,
```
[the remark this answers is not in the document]
tobyhede: I think this is a lot of code for a warning that could be handled by:

```
if let Ok(value) = std::env::var(CS_AUTH__WORKSPACE_ID)) {
  warn!(
}
```

Region has been added just to be made obsolete.

tobyhede (approved):

packages/cipherstash-proxy/src/config/tandem.rs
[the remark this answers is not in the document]
yujiyokoo: I did it this way because checking the env var alone will not pick up the workspace IDs in toml files. Not sure who or if anyone has them besides me but I did. Similarly, customers might not have regions but I did in my toml files so I added it too. I think we should delete this code in a while.

Cargo.lock
yujiyokoo: Not exactly sure what you mean here. It's set to this version in Cargo.toml, unless you mean something more than that

PR #253, opened 2025-06-02
Ensure schema column names distinct

tobyhede:
Column names are being duplicated in the schema because of the join to find primary key. 
Multiple columns may be the "Primary key" as postgresql allows composite keys.
The primary key is not used by the mapper at all, so the fix is to remove the join and remove primary key from the schema config.

### DIFF
--- a/packages/cipherstash-proxy-integration/src/select/mod.rs
+++ b/packages/cipherstash-proxy-integration/src/select/mod.rs
@@ -2,3 +2,4 @@ mod group_by;
 mod order_by;
 mod order_by_with_null;
 mod pg_catalog;
+mod unmappable;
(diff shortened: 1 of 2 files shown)


packages/eql-mapper/src/model/schema_delta.rs
```diff
@@ -141,13 +139,7 @@ impl OverlayTable {
...
     fn remove_column(&mut self, name: &Ident) {
-        if let Some((idx, _)) = self
-            .columns
-            .iter()
-            .enumerate()
-            .find(|(_, col)| col.name == *name)
-        {
-            self.primary_key.retain(|col| *col != idx);
+        if self.columns.iter().any(|col| col.name == *name) {
             self.columns.retain(|col| col.name != *name);
```
tobyhede, line 143: @freshtonic Do you remember the intention behind retain rather than remove

PR #257, opened 2025-06-03
Feature/elixir integration test

yujiyokoo:
This adds a few Elixir integration tests using Ecto and Postgrex. What it does is very similar to the tests we have in golang.

A lot of the files are generated using the `mix` command and modified.

One of the tests is for the currently unsupported query, and expects an error (Ecto turns literal `7.5` into `7.5::float` in the final SQL).

## Acknowledgment


tests/integration/elixir_test/test/elixir_test_test.exs
```diff
@@ -0,0 +1,201 @@
...
+
+    # Ecto appends explicit cast to `7.5`, making it `7.5::float` and causes
+    # the "operator does not exist" error
+    q =
+      from(e in "encrypted",
+        where: e.encrypted_float8 == 7.5,
+        select: [e.id, e.encrypted_float8]
+      )
+
+    assert_raise(Postgrex.Error, fn -> Repo.all(q) end)
```
tobyhede, line 120: 👍

PR #260, opened 2025-06-10
chore: enable language-specific tests in test:integration

yujiyokoo:
It looks like golang and python weren't present in mise.toml. Enabling all.
## Acknowledgment


auxesis (approved):
Thanks for this @yujiyokoo!

PR #264, opened 2025-06-16
chore: fix all breakages due to type changes in sqlparser

freshtonic:
This was a *mission*.

The main culprits were `ObjectName(Vec<Ident>)` changing to `ObjectName(Vec<ObjectNamePart>)` and `Expr::Value(Value)` changing to `Expr::Value(ValueWithSpan)`.

## Acknowledgment


tobyhede (approved):
Is very good to have types.

PR #273, opened 2025-07-04
Tests for SQL UPDATE

tobyhede:
UPDATE tests following the implementation patterns of the INSERT tests.

Collects all of the update tests into a single update module, covering both simple and extended protocols


packages/cipherstash-proxy-integration/src/update/update_domain_type.rs
```diff
@@ -0,0 +1,76 @@
...
+    #[derive(Debug, ToSql, FromSql, PartialEq)]
+    #[postgres(name = "domain_type_with_check")]
+    pub struct Domain(String);
+
+    ///
+    /// Tests update of custom domain type
+    ///
+    #[tokio::test]
+    async fn update_domain_type() {
+        trace();
```
coderdan, line 15: Handy!

coderdan (approved):

PR #275, opened 2025-07-08
Tests for JSONB field access operator (`->`)

tobyhede:
Adds tests for JSONB field access operators (`->`)

Converts field names into a JSONPath selector for compatibility with `cipherstash-client`.


auxesis (approved):
❤️ nice work @tobyhede.

PR #280, opened 2025-07-16
fix(eql-mapper): resolution of CTE tables

freshtonic:
Resolutions of CTE tables would only succeed from the main statement of a `WITH` but not sub-statements of that statement.

The fix was to not create parentless scopes on entry to Statement nodes - instead create a new scope pointing to the parent scope.

## Acknowledgment


auxesis (approved):
LGTM, thanks @freshtonic.

PR #290, opened 2025-07-24
fix: error on invalid workspace_crn

tobyhede:
crn parse error were being ignored - the variable was not set, leading to a "missing config" error message.

This change explictly handles parse errors caused by invalid `workspace_crn` and outputs an appropriate message

### DIFF
--- a/packages/cipherstash-proxy/src/config/tandem.rs
+++ b/packages/cipherstash-proxy/src/config/tandem.rs
@@ -138,6 +138,11 @@ pub struct DevelopmentConfig {
 ///
 /// ENV vars must be prefixed with `CS_`.
 ///
+/// IMPORTANT!
+/// Config needs to be loaded *before* logging can be initalized
+/// Tracing macros will not output and you may be confused.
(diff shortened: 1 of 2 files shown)


packages/cipherstash-proxy/src/config/tandem.rs
```diff
@@ -138,6 +138,11 @@ pub struct DevelopmentConfig {
 ///
 /// ENV vars must be prefixed with `CS_`.
 ///
+/// IMPORTANT!
+/// Config needs to be loaded *before* logging can be initalized
+/// Tracing macros will not output and you may be confused.
+/// Use `println!`` and `eprintln!`` instead.
+///
```
freshtonic, line 145: Nice

freshtonic (approved):

packages/cipherstash-proxy/src/config/tandem.rs
tobyhede: Yeah, I confused myself for quite a while. 🫠

PR #297, opened 2025-08-01
chore: 🔧 disable elixir integration tests

freshtonic:
Comment out elixir integration test execution in mise.toml task configuration.

## Acknowledgment


mise.toml
```diff
@@ -351,7 +351,7 @@ echo '###############################################'
 echo
 mise run test:integration:lang:golang
 mise run test:integration:lang:python
-mise run test:integration:lang:elixir
+# mise run test:integration:lang:elixir
```
yujiyokoo, line 354: I think it might help to add a comment why this is commented out here in case it takes a while to get back to this issue but I'm hoping that won't be the case

yujiyokoo (approved):
LGTM. Minor comment but no blocker

PR #299, opened 2025-08-01
Add SET command for run-time configuration of keyset_id

tobyhede:
The Proxy should expose a `SET` command that enables run-time configuration of a `keyset_id`: 

```
SET CIPHERSTASH.KEYSET_ID = 'EXAMPLE-UUID-4b38-9571-e0cf47802677'
```

- [x] Parse set command
- [x] Parse keyset as UUID if defined  
- [x] Set keyset_id if defined
- [x] Error if set command but value cannot be parsed 
- [x] Update error docs with syntax details 
- [x] Extract zerokms initalisation 
- [x] Enable cipher initalisation to accept keyset_id 
- [x] Pass keyset_id when encrypting or decrypting
- [x] Handle decrypt errors
- [x] Refactor to extract error handling from frontend into common module
- [x] Update backend to send decrypt errors as pg errors


freshtonic:
This looks really good - nice work

freshtonic (approved):
Stellar work, Mr. Hede

PR #304, opened 2025-08-12
Set keyset by name

tobyhede:
- [x] `SET KEYSET` statement parsing 
- [x] Creates a `KeysetIdentifier` that can handle name or id. 
- [ ] Integration tests (requires updates in client library first)

### DIFF
--- a/packages/cipherstash-proxy-integration/src/lib.rs
+++ b/packages/cipherstash-proxy-integration/src/lib.rs
@@ -18,6 +18,7 @@ mod pipeline;
 mod schema_change;
 mod select;
 mod set_keyset_id;
+mod set_keyset_name;
 mod simple_protocol;
 mod support;
 mod update;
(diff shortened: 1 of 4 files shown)


packages/cipherstash-proxy/src/config/tandem.rs
```diff
@@ -176,12 +176,14 @@ impl TandemConfig {
         // For parsing top-level values such as CS_HOST, CS_PORT
         // and for parsing nested env values such as CS_DATABASE__HOST, CS_DATABASE__PORT
         let cs_env_source = Environment::with_prefix(CS_PREFIX)
+            .ignore_empty(true)
```
tobyhede, line 179: docker-compose will pass an empty value if an `ENV` var is not set.
Proxy will now ignore any empty values.

freshtonic (approved):
Amazing.

How did we get this far without clearing errors and sending `ReadyForQuery` though?

PR #306, opened 2025-08-20
Set key name integration test

tobyhede:
- [ ] SET KEYSET_NAME tests
- [x] Fix integration test output
- [x] Reduce proxy logging in CI
- [x] Use blacksmith runner

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -31,6 +31,8 @@ jobs:
           CS_CLIENT_ID: ${{ secrets.CS_CLIENT_ID }}
           CS_CLIENT_KEY: ${{ secrets.CS_CLIENT_KEY }}
           CS_WORKSPACE_CRN: "crn:ap-southeast-2.aws:${{ secrets.CS_WORKSPACE_ID }}"
+          CS_LOG__FORMAT: "structured"
+          CS_LOG__LEVEL: "info"
 
           RUST_BACKTRACE: "1"
         run: |
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -14,7 +14,7 @@ env:
 jobs:
   test:
     name: Test
-    runs-on: buildjet-16vcpu-ubuntu-2204
+    runs-on: blacksmith-8vcpu-ubuntu-2404
     steps:
       - uses: actions/checkout@v4
       - uses: ./.github/actions/setup-test
--- a/mise.toml
+++ b/mise.toml
@@ -278,20 +278,20 @@ description = "Runs integration test/s"
 run = """
 set -e
 
-# echo
-# echo '###############################################'
-# echo '# Preflight'
-# echo '###############################################'
-# echo
+echo
+echo '###############################################'
+echo '# Preflight'
+echo '###############################################'
+echo
 
 # Ensure Postgres instances are running
 mise run test:integration:preflight
 
-# echo
-# echo '###############################################'
-# echo '# Test: unconfigured proxy'
-# echo '###############################################'
-# echo
+echo
+echo '###############################################'
+echo '# Test: unconfigured proxy'
+echo '###############################################'
+echo
 
 mise --env tcp run postgres:eql:teardown
 mise --env tcp run proxy:up proxy --extra-args "--detach --wait"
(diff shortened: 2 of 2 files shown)


packages/cipherstash-proxy-integration/src/multitenant/set_keyset_name.rs
```diff
@@ -0,0 +1,374 @@
...
+        assert!(result.is_ok());
+
+        // SELECT TENANT_2 record with TENANT_1 should fail
+        let sql = format!("SELECT id, encrypted_text FROM encrypted WHERE id = {tenant_2_id}");
+        let result = tenant_1_client.simple_query(&sql).await;
+        assert!(result.is_err());
+
+        // ------------------
+
+        // SELECT TENANT_2 record with TENANT_1 should fail
```
yujiyokoo, line 223: This comment seems incorrect

packages/cipherstash-proxy-integration/src/multitenant/set_keyset_name.rs
```diff
@@ -0,0 +1,374 @@
...
+
+        // TENANT_2 INSERT
+        let result = tenant_2_client
+            .query(insert_sql, &[&tenant_2_id, &tenant_2_text])
+            .await;
+        assert!(result.is_ok());
+
+        // ------------------
+
+        // SELECT TENANT_1 record with TENANT_2 should fail
```
yujiyokoo, line 211: I think comment should be 'succeed' instead of 'fail'

packages/cipherstash-proxy-integration/src/multitenant/set_keyset_name.rs
```diff
@@ -0,0 +1,374 @@
...
+
+        // DEFAULT_KEYSET_ID SHOULD BE DISABLED FOR THIS TEST
+        // SET KEYSET IS REQUIRED
+        let sql = "INSERT INTO encrypted (id, encrypted_text) VALUES (1, 'hello')";
+        let result = client.simple_query(sql).await;
+        assert!(result.is_err());
+
+        // // THIS IS COUNTER INTUITIVE
+        // // Statement passes through frontend because no mapping is required
+        // // Returns without error because there is NO DATA and decrypt is never called
```
yujiyokoo, line 99: Are these double comments left over from commenting-and-uncommenting in bulk here (and below)?

packages/cipherstash-proxy-integration/src/multitenant/set_keyset_id.rs
```diff
@@ -246,6 +246,44 @@ mod tests {
...
+        let result = client.query(sql, &[]).await;
+        assert!(result.is_ok());
+
+        // Error on encrypt
+        let id = random_id();
+        let text = "TEST UNKNOWN";
+
+        let insert_sql = "INSERT INTO encrypted (id, encrypted_text) VALUES ($1, $2)";
+        let result = client.query(insert_sql, &[&id, &text]).await;
+        assert!(result.is_err());
```
yujiyokoo, line 274: I think this is fine, but just out of curiosity, is it possible/worthwhile to check the error types?

packages/cipherstash-proxy-integration/src/multitenant/set_keyset_name.rs
```diff
@@ -0,0 +1,374 @@
...
+        //  --------
+        // INSERT and SELECT as TENANT 1
+        // SET tenant 1
+        let sql = format!("SET CIPHERSTASH.KEYSET_NAME = '{tenant_keyset_name_1}'");
+        let result = client.query(&sql, &[]).await;
+        assert!(result.is_ok(), "Setting keyset_name should succeed");
+
+        // INSERT
+        let tenant_1_id = random_id();
+        let encrypted_val = 2;
```
yujiyokoo, line 40: nit: I found this name slightly confusing, because I thought '2 looks like a plaintext value', but I suppose this is named like this as it's 'value to be encrypted'

yujiyokoo (approved):
Looks good. Added some questions and comments about comments but nothing blocking

packages/cipherstash-proxy-integration/src/multitenant/set_keyset_id.rs
tobyhede: I was thinking the same thing after I pushed this, actually. 
And probably worth extending across the other error tests.

tobyhede:
@yujiyokoo Fixed issues and added tests for the 'Unknown keyset" error message

PR #308, opened 2025-08-20
feat: cache initialised scoped_cipher/s

tobyhede:
The client scoped_cipher fetches keyset information on init.

In multitenant operation, many keysets are used, and the initialisation occurs on every encrypt or decrypt operation.

Once initialised, scoped_ciphers should be retained to avoid the cost of the initial request for every operation.


packages/cipherstash-proxy/src/encrypt/mod.rs
```diff
@@ -82,28 +86,81 @@ impl Encrypt {
...
 
+        let cipher_cache = Cache::builder()
+            // Use weigher to estimate memory usage of ScopedCipher instances
+            .weigher(|_key: &String, _value: &Arc<ScopedCipher>| -> u32 {
+                // Estimate memory usage of a ScopedCipher:
+                // - Base Arc overhead: ~24 bytes
+                // - ScopedCipher internal state: estimated ~1KB for crypto state
+                // - Key material and connection state: estimated ~2KB
+                // Total conservative estimate: ~3KB per cipher
+                3 * 1024 // 3KB estimated per cipher
```
freshtonic, line 97: You can do this accurately with `std::mem::size_of<T>()`.

See: https://doc.rust-lang.org/std/mem/fn.size_of.html

freshtonic (approved):
Approved, but see my comment about using `std::mem::size_of`

PR #309, opened 2025-08-21
refactor: ♻️ use std::mem::size_of for accurate cipher cache sizing

tobyhede:
Replace hardcoded 3KB estimate with actual ScopedCipher struct size calculation:
- Add SCOPED_CIPHER_SIZE constant using std::mem::size_of
- Update cache weigher to use accurate size (~56 bytes vs 3KB estimate)
- Update max_capacity calculation to match weigher
- Improves cache efficiency by allowing ~50x more entries with same memory footprint

### DIFF
--- a/docs/reference/index.md
+++ b/docs/reference/index.md
@@ -258,6 +258,55 @@ As a convenience for production deployments, with the below environment variable
 CS_DATABASE__INSTALL_AWS_RDS_CERT_BUNDLE="true"
 ```
 
+## Multitenant
+
+CipherStash Proxy supports multitenant applications by allowing clients to switch between different keysets at runtime. This enables a single Proxy instance to handle encrypted data for multiple tenants, with each tenant's data protected by separate encryption keys.
+
+### Keyset Commands
+
+#### SET CIPHERSTASH.KEYSET_ID
+
+Sets the active keyset for the current connection using a keyset UUID.
+
+**Syntax:**
+```sql
+SET CIPHERSTASH.KEYSET_ID = '<keyset-uuid>';
+```
+
+**Parameters:**
+- `keyset-uuid`: The UUID of the keyset to activate for this connection
+
+**Example:**
+```sql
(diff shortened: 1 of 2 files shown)


docs/reference/index.md
```diff
@@ -258,6 +258,55 @@ As a convenience for production deployments, with the below environment variable
 CS_DATABASE__INSTALL_AWS_RDS_CERT_BUNDLE="true"
 ```
 
+## Multitenant
```
yujiyokoo, line 261: Should it be "Multitenancy" or perhaps something like "Multitenant mode"?

docs/reference/index.md
```diff
@@ -258,6 +258,55 @@ As a convenience for production deployments, with the below environment variable
...
+- `keyset-name`: The name of the keyset to activate for this connection
+
+**Example:**
+```sql
+SET CIPHERSTASH.KEYSET_NAME = 'tenant-1';
+```
+
+### Usage Notes
+
+- These commands must be executed before performing any encrypted operations
```
yujiyokoo, line 303: Not sure, but would "encryption" be more suitable than "encrypted" here?

docs/reference/index.md
```diff
@@ -258,6 +258,55 @@ As a convenience for production deployments, with the below environment variable
 CS_DATABASE__INSTALL_AWS_RDS_CERT_BUNDLE="true"
 ```
 
+## Multitenant
+
+CipherStash Proxy supports multitenant applications by allowing clients to switch between different keysets at runtime. This enables a single Proxy instance to handle encrypted data for multiple tenants, with each tenant's data protected by separate encryption keys.
+
+### Keyset Commands
```
yujiyokoo, line 265: It may be helpful to the users if we tell them whether they need to set both or one of those

yujiyokoo (approved):
LGTM. A couple of non-blocking comments

docs/reference/index.md
tobyhede: Good call.

PR #314, opened 2025-08-27
♻️ refactor: convert Name struct to enum and extract to own module

tobyhede:
♻️ refactor: convert Name struct to enum and extract to own module
- Convert Name from struct to enum with Named(String) and Unnamed variants
- Extract Name type into dedicated name.rs module for better organization
- Add From<String> and From<&str> implementations for convenient construction
- Update all usage sites to use Name::from() constructor
- Update field access from .0 to .as_str() method
- Maintain backward compatibility through re-exports and Deref trait
- Improve type safety by preventing empty named statements

### DIFF
--- a/packages/cipherstash-proxy/src/postgresql/messages/close.rs
+++ b/packages/cipherstash-proxy/src/postgresql/messages/close.rs
@@ -52,7 +52,7 @@ impl TryFrom<&BytesMut> for Close {
         let target = cursor.get_u8();
         let target = Target::try_from(target)?;
         let name = cursor.read_string()?;
-        let name = Name(name);
+        let name = Name::from(name);
 
         Ok(Close { target, name })
     }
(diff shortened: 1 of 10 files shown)


packages/cipherstash-proxy/src/postgresql/messages/close.rs
```diff
@@ -0,0 +1,144 @@
...
+impl TryFrom<Close> for BytesMut {
+    type Error = Error;
+
+    fn try_from(close: Close) -> Result<BytesMut, Error> {
+        let mut bytes = BytesMut::new();
+
+        let name = CString::new(close.name.as_str())?;
+        let name = name.as_bytes_with_nul();
+
+        let len = SIZE_I32 + SIZE_U8 + name.len();
```
freshtonic, line 70: Rust has `std::mem::size_of<T>()` to return sizes of types in bytes.

Consider using `size_of::<i32>() + size_of::<u8>() + name.len()` which is more idiomatic than defining our own constants primitive type sizes.

freshtonic (approved):
LGTM but I suggest using `std::mem::size_of` instead of the crate-defined constants for byte-sizes of primitives.

packages/cipherstash-proxy/src/postgresql/messages/close.rs
tobyhede: @freshtonic The values are used in a few places and are defined using size_of:

```
pub const SIZE_U8: usize = mem::size_of::<u8>();
pub const SIZE_I16: usize = mem::size_of::<i16>();
pub const SIZE_I32: usize = mem::size_of::<i32>();
```

PR #316, opened 2025-08-27
refactor: centralize log target definitions using declarative macro

tobyhede:
Replace manual log target management with a declarative macro system that defines all targets in one place. This eliminates the need to update 4-5 places when adding new targets and ensures consistency.

Key changes:
- Create log/targets.rs with define_log_targets! macro
- Generate target constants, functions, and level mappings automatically
- Update LogConfig to use centrally-defined field list
- Maintain all existing functionality and public API

Benefits:
- Single source of truth for log targets
- Compile-time safety between targets and config fields
- Self-documenting and easier to maintain
- Eliminates repetitive boilerplate code

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -244,17 +244,23 @@ Default level is `Info`.
 
 Proxy has multiple "log targets" corresponding to the internal domains.
 
-Set log levels for a specific log target to turn on or turn of more verbose logging:
+Set log levels for a specific log target to turn on or turn off more verbose logging:
 
 ```
 Target          | ENV
 --------------- | -------------------------------------
 DEVELOPMENT     | CS_LOG__DEVELOPMENT_LEVEL
 AUTHENTICATION  | CS_LOG__AUTHENTICATION_LEVEL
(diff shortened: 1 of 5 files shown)


freshtonic (approved):
I'm fine with it but I get a mid-confidence vibe that we're not using `tracing-subscriber` idiomatically as this feels more complex than it should be (not the changes in this PR per se because it was more verbose before this PR).

freshtonic:
@tobyhede I will try to articulate my vibe-based concerns better :)

The macro reduces boilerplate which is a good thing.

My concern is that both before and after the introduction of the macro we were possibly holding `tracing` and `tracing-subscriber` wrong.

Definition of targets (and what is emitted) is the responsibility of the log _publisher_. The extent to which the publisher should concern itself with log _levels_ should be limited to choosing a the log macro to use e.g. `debug!` / `info!` / `trace!` etc.

Definition of active log levels per target is a _subscriber_ concern. 

We've tied together the definition of a target (a publisher-defined + subscribers-uses) with a log level default - which is a subscriber concern.

1. It's a feeling
2. I could be wrong about it
3. Not a hill I'll die on & it should not block merging the PR

PR #327, opened 2025-09-12
Refactor frontend context

tobyhede:
Incremental refactoring Part `1` of `N`

The concept of the plan is to remove all the encryption and other business logic from the frontend and backend handlers.
The `context` will become the coordinator, responsible for parsing, type-checking, mapping and encrypting data, as well as maintaining session state. 

This PR makes context the keeper of all of the data required for message handling: 
- proxy config, 
- encrypt config
- schema
- zerokms

The schema reloading has also changed.  
Rather than pass the entire schema service so the reload function can be called, uses tokio channels. 
We have not implemented a mechanism to trigger the encrypt config reloading, but will use the same channel mechanism and the command has been implemented. 

Context is now the primary interface for handler operations
All handler dependencies now flow through Context

Remaining work:
- context is now clearly too big
- some functions in frontend still need to be extracted.
 
Plan is to get the functions out of the frontend and iterate on new abstractions within the context boundary. 

Further work will be pushed in separate PRs.

### DIFF
--- a/docs/errors.md
+++ b/docs/errors.md
@@ -32,6 +32,8 @@
 
 - Configuration errors:
   - [Missing or invalid TLS configuration](#config-missing-or-invalid-tls)
+  - [Network configuration change requires restart](#config-network-change-requires-restart)
+
 <!-- ---------------------------------------------------------------------------------------------------- -->
 
 <!-- ---------------------------------------------------------------------------------------------------- -->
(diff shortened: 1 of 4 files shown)


packages/cipherstash-proxy/src/config/database.rs
```diff
@@ -81,6 +81,21 @@ impl DatabaseConfig {
         })?;
         Ok(name)
     }
+
+    #[cfg(test)]
+    pub fn for_testing() -> Self {
```
tobyhede, line 86: Started with  `impl Default` but a `default` implies that there is a sensible default, but there is not.

Went with`for_testing` instead, but open to ideas.

packages/cipherstash-proxy/src/postgresql/column_mapper.rs
```diff
@@ -0,0 +1,167 @@
...
+use tracing::{debug, warn};
+
+/// Service responsible for processing columns from type-checked SQL statements
+/// and mapping them to encryption configurations.
+#[derive(Clone)]
+pub struct ColumnMapper {
+    encrypt_config: Arc<EncryptConfig>,
+}
+
+impl ColumnMapper {
```
tobyhede, line 20: Functions responsible for mapping a column to configuration lifted straight from frontend.

I expect this to evolve with further refactoring.

packages/cipherstash-proxy/src/proxy/mod.rs
```diff
@@ -43,96 +61,109 @@ impl Proxy {
...
+                    msg = "Could not query EQL version from database",
+                    error = err.to_string()
+                );
+                None
+            }
+        };
+        Ok(version)
+    }
+
+    pub fn receive(
```
tobyhede, line 107: Responsible for receiving `ReloadCommand` messages. 

The `ReloadCommand` holds a `responder` that lets the calling context know that reload is complete.

freshtonic (approved):
Factoring<sup>after</sup> > Factoring<sup>before</sup> → _approved_

Added a few comments.

PR #328, opened 2025-09-29
feat(config): add optional database connection CLI args

coderdan:
Adds optional command line args to proxy for specifying the database, host and username.
If not provided, env and config is used as now.
If running in docker this has no effect but is very useful if running the binary directly.

## Acknowledgment


auxesis (commented):
Love this, thanks for the PR @coderdan — big usability improvement. 

Could you add something to `docs/reference/index.md` explaining how to start Proxy with these new command line options?

Looking forward to taking this for a spin. ❤️

auxesis (approved):

PR #385, opened 2026-03-25
Docs/sync cipherstash client 0.34

tobyhede:
## Acknowledgment


docs/reference/searchable-json.md
```diff
@@ -302,7 +302,7 @@ SELECT encrypted_jsonb @> '{"number": 99}' FROM cipherstash;
 
 ```sql
 -- nested object
-SELECT encrypted_jsonb @> '{"object": {"string": "world", "number": 99},' FROM cipherstash;
+SELECT encrypted_jsonb @> '{"object": {"string": "world", "number": 99}}' FROM cipherstash;
```
tobyhede: Correctness fix. 

In PostgreSQL:

  - @> means "left contains right" — column @> '{"number": 99}'
  - <@ means "left is contained by right" — '{"number": 99}' <@ column

  The old version had the operands flipped:

  -- WRONG: asks "does the literal contain the column?" (backwards)
  SELECT '{"number": 99}' @> encrypted_jsonb FROM cipherstash;

  -- CORRECT: asks "is the literal contained by the column?"
  SELECT '{"number": 99}' <@ encrypted_jsonb FROM cipherstash;

  When the literal is on the left side, you need <@ (contained by), not @> (contains). The old version
  had the operator semantics inverted relative to the operand order. Both queries might have returned
  the same result in practice (containment is symmetric for simple key-value pairs at top level), but
  the operator was semantically wrong for how the operands were positioned.

PR #390, opened 2026-04-27
feat(proxy): support OPE index type

freshtonic:
## Summary

Adds the `ope` (Order-Preserving Encryption) index type as a drop-in alternative to `ore` for range and ordering queries on encrypted columns. OPE ciphertexts compare under standard lexicographic byte ordering, so they support the same operator set as ORE (`<`, `<=`, `>`, `>=`, plus `=` on non-text types) but with cheaper comparisons.

Configure with:

```sql
SELECT eql_v2.add_search_config('users', 'salary', 'ope', 'int');
```

Anywhere `ore` works for range or `ORDER BY`, `ope` works the same way. Pick one per column, not both.

## How it's wired

`IndexType::Ope`, `OpeIndexOpts`, and `Index::new_ope()` all live upstream in `cipherstash-config` (already a workspace dep at `0.34.1-alpha.4`). `eql-mapper` doesn't branch on `IndexType`, so OPE flows through transparently. This PR therefore mainly:

- Adds integration coverage that exercises the wire-up end-to-end against a real proxy/postgres pair.
- Adds a parser unit test that asserts `"ope": {}` deserialises to `IndexType::Ope`.
- Documents OPE as a configuration option.

### Tests (`packages/cipherstash-proxy-integration`, `tests/sql/schema.sql`)

- New `map_ope_index_where` (7 tests, one per cast type: int2/int4/int8, float8, date, text, bool) and `map_ope_index_order` (6 tests covering `ASC`/`DESC` and `NULLS FIRST`/`NULLS LAST`).
- Per-test fixture tables for both ORE and OPE generated by a `DO` block in `schema.sql`, eliminating parallel-test races that previously required `#[serial]`.
- `ore_order_helpers` parameterised on table name so the same helpers serve per-test ORE tables and the shared multitenant `encrypted` table.

End-to-end (parallel, default thread count): 38/38 ORE+OPE tests pass in ~1.6s. Previously 14s serial / flaky in parallel.

### Unit test (`packages/cipherstash-proxy/src/proxy/encrypt_config/manager.rs`)

`can_parse_ope_index` mirroring `can_parse_ore_index` — asserts the canonical config parser accepts `{"indexes": {"ope": {}}}` and produces `IndexType::Ope`.

### Docs

- `docs/how-to/index.md` — adds `ope` to the list of available index types alongside `match` and `ore`, with a note that `ope` and `ore` are mutually exclusive per column.
- `docs/reference/searchable-json.md` — `MIN()`/`MAX()` line now reads `with an 'ore' or 'ope' index`.
- `docs/sql/schema-example.sql` — comment near the `ore` entry pointing to `ope` as the alternative.
- `CHANGELOG.md` — `[Unreleased]` entry under `Added`.

## Test plan

- [x] `cargo nextest run -p cipherstash-proxy -E 'test(can_parse)'` — 9/9 pass (includes new `can_parse_ope_index`).
- [x] `cargo nextest run -p cipherstash-proxy-integration -E 'test(map_ope_index)'` — 13/13 pass.
- [x] `cargo nextest run -p cipherstash-proxy-integration -E 'test(map_ore_index)'` — 25/25 pass (regression check after fixture refactor).
- [x] `cargo fmt --check` and `cargo clippy --tests -- -D warnings` clean on `cipherstash-proxy`.

## Acknowledgment

## Summary by CodeRabbit

* **New Features**
  * Introduced OPE (Order-Preserving Encryption) index for range and ORDER BY queries as a drop‑in alternative to the existing range index.

* **Documentation**
  * Updated examples and guidance to explain OPE vs other encrypted index types and clarified MIN()/MAX() support.

* **Tests**
  * Added/expanded unit and integration tests covering OPE ordering and WHERE semantics, per-test fixture tables, and table-cleanup helpers.

* **Chores**
  * Added CHANGELOG entry documenting the new index.

[Review Change Stack](https://app.coderabbit.ai/change-stack/cipherstash/proxy/pull/390)

### DIFF
--- a/packages/cipherstash-proxy-integration/src/map_ope_index_where.rs
+++ b/packages/cipherstash-proxy-integration/src/map_ope_index_where.rs
@@ -1,0 +1,173 @@
+#[cfg(test)]
+mod tests {
+    use crate::common::{clear_ope, connect_with_tls, random_id, trace, PROXY};
+    use chrono::NaiveDate;
+    use tokio_postgres::types::{FromSql, ToSql};
+    use tokio_postgres::Client;
+
+    #[tokio::test]
+    async fn map_ope_where_generic_int2() {
+        map_ope_where_generic("encrypted_int2", 40i16, 99i16).await;
+    }
+
+    #[tokio::test]
+    async fn map_ope_where_generic_int4() {
+        map_ope_where_generic("encrypted_int4", 40i32, 99i32).await;
+    }
+
+    #[tokio::test]
+    async fn map_ope_where_generic_int8() {
+        map_ope_where_generic("encrypted_int8", 40i64, 99i64).await;
+    }
+
+    #[tokio::test]
+    async fn map_ope_where_generic_float8() {
+        map_ope_where_generic("encrypted_float8", 40.0f64, 99.0f64).await;
+    }
+
+    #[tokio::test]
+    async fn map_ope_where_generic_date() {
+        let low = NaiveDate::parse_from_str("2024-01-01", "%Y-%m-%d").unwrap();
+        let high = NaiveDate::parse_from_str("2027-01-01", "%Y-%m-%d").unwrap();
+        map_ope_where_generic("encrypted_date", low, high).await;
+    }
+
+    #[tokio::test]
+    async fn map_ope_where_generic_text() {
+        map_ope_where_generic("encrypted_text", "ABC".to_string(), "BCD".to_string()).await;
+    }
+
+    #[tokio::test]
+    async fn map_ope_where_generic_bool() {
+        map_ope_where_generic("encrypted_bool", false, true).await;
+    }
+
+    /// Tests OPE operations on the `encrypted_ope` table with 2 values - high & low.
+    /// Mirrors `map_ore_where_generic` but targets the OPE-indexed mirror table.
+    async fn map_ope_where_generic<T>(col_name: &str, low: T, high: T)
+    where
+        for<'a> T: Clone + PartialEq + ToSql + Sync + FromSql<'a> + PartialOrd,
+    {
+        trace();
+
+        clear_ope().await;
+
+        let client = connect_with_tls(PROXY).await;
+
+        // Insert test data
+        let sql = format!("INSERT INTO encrypted_ope (id, {col_name}) VALUES ($1, $2)");
+        for val in [low.clone(), high.clone()] {
+            client
+                .query(&sql, &[&random_id(), &val])
+                .await
+                .expect("insert failed");
+        }
+
+        // NULL record
+        let sql = format!("INSERT INTO encrypted_ope (id, {col_name}) VALUES ($1, null)");
+        client
+            .query(&sql, &[&random_id()])
+            .await
+            .expect("insert failed");
+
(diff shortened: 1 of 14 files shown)


packages/cipherstash-proxy-integration/src/map_ope_index_where.rs
```diff
@@ -0,0 +1,208 @@
...
+    #[tokio::test]
+    async fn map_ope_where_generic_text() {
+        map_ope_where_generic(
+            "encrypted_ope_where_text",
+            "encrypted_text",
+            "ABC".to_string(),
+            "BCD".to_string(),
+        )
+        .await;
+    }
```
coderdan, line 52: We should test text with differing length strings as well.

packages/cipherstash-proxy-integration/src/map_ore_index_where.rs
```diff
@@ -1,62 +1,97 @@
...
     }
 
     #[tokio::test]
     async fn map_ore_where_generic_text() {
-        map_ore_where_generic("encrypted_text", "ABC".to_string(), "BCD".to_string()).await;
+        map_ore_where_generic(
+            "encrypted_ore_where_text",
+            "encrypted_text",
+            "ABC".to_string(),
+            "BCD".to_string(),
```
coderdan: As above, we should test strings of differing lengths as well.

tests/sql/schema.sql
```diff
@@ -169,6 +169,144 @@ SELECT eql_v2.add_search_config(
...
+        id bigint,
+        plaintext text,
+        plaintext_date date,
+        encrypted_text eql_v2_encrypted,
+        encrypted_bool eql_v2_encrypted,
+        encrypted_int2 eql_v2_encrypted,
+        encrypted_int4 eql_v2_encrypted,
+        encrypted_int8 eql_v2_encrypted,
+        encrypted_float8 eql_v2_encrypted,
+        encrypted_date eql_v2_encrypted,
```
coderdan: Is this adding all columns for all tables? Might be a bit redundant if so. Not a major problem.

Cargo.toml
```diff
@@ -43,8 +43,11 @@ debug = true
 
 [workspace.dependencies]
 sqltk = { version = "0.10.0" }
-cipherstash-client = { version = "0.34.0-alpha.4" }
-cts-common = { version = "0.34.0-alpha.4" }
+# TODO: revert to crates.io once the cipherstash-suite version with `IndexType::Ope`
+# is published. See https://github.com/cipherstash/cipherstash-suite for the
+# OPE-enabled change.
+cipherstash-client = { path = "../cipherstash-suite/packages/cipherstash-client" }
+cts-common = { path = "../cipherstash-suite/packages/cts-common" }
```
coderdan: I'm assuming these will be updated before merge.

coderdan (approved):
Some minor suggestions.
Only blocker is to update the peer-dependencies for cipherstash-suite.

coderdan (approved):
Some small suggestions. LGTM

tests/sql/schema.sql
freshtonic: Yes it is, but it's following an existing pattern which we can change in a follow up.

PR #391, opened 2026-04-27
fix(proxy): use specific `EqlTrait`s on encrypted columns (instead of `EqlTrait::all()`)

freshtonic:
## Acknowledgment

### DIFF
--- a/packages/cipherstash-proxy/src/proxy/mod.rs
+++ b/packages/cipherstash-proxy/src/proxy/mod.rs
@@ -63,7 +63,8 @@ impl Proxy {
 
         let encrypt_config_manager = EncryptConfigManager::init(&config.database).await?;
 
-        let schema_manager = SchemaManager::init(&config.database).await?;
+        let schema_manager =
+            SchemaManager::init(&config.database, encrypt_config_manager.clone()).await?;
 
         let eql_version = Proxy::eql_version(&config).await?;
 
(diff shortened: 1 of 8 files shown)


tobyhede (approved):

packages/cipherstash-proxy/src/proxy/schema/manager.rs
```diff
@@ -141,9 +164,18 @@ pub async fn load_schema(config: &DatabaseConfig) -> Result<Schema, Error> {
...
             let column = match column_type_name.as_deref() {
                 Some("eql_v2_encrypted") => {
-                    debug!(target: SCHEMA, msg = "eql_v2_encrypted column", table = table_name, column = col);
-
-                    let eql_traits =  EqlTraits::all();
+                    let identifier = Identifier::new(&table_name, col);
+                    let eql_traits = encrypt_config
+                        .get_column_config(&identifier)
+                        .map(|config| eql_traits_from_indexes(&config.indexes))
+                        .unwrap_or_default();
```
tobyhede, line 171: Encrypted without config has silent fallback.
Log may be helpful here

```
  .unwrap_or_else(|| {
      warn!(
          target: SCHEMA,
          msg = "eql_v2_encrypted column has no encrypt_config entry; refusing all encrypted operations",
          table = table_name,
          column = col,
      );
      EqlTraits::none()
  });
```

packages/cipherstash-proxy/src/proxy/mod.rs
tobyhede, line 126: The `ReloadCommand::EncryptSchema` should probably refresh the database schema now the calls are dependent. 

```
  ReloadCommand::EncryptSchema(responder) => {
      encrypt_config_manager.reload().await;
      schema_manager.reload().await;
      let _ = responder.send(());
  }
```

tobyhede (approved):

PR #395, opened 2026-05-29
fix(passthrough): drain per-statement queues to stop memory leak

### LINKED ISSUE #400: Memory leak in passthrough mode (empty encrypt config) — per-statement leak causes OOM
## Summary

Proxy leaks memory **per statement** when running in passthrough with an **empty encrypt config** (a deployment where encryption has not been configured). RSS climbs steadily until the pod is OOM-killed. Configured deployments are unaffected.

**Present in both 2.1.23 and 2.2.1 (latest) — 2.2.1 is worse. Not fixed in any current release.**

## Workaround (no upgrade required)

Set `CS_DEVELOPMENT__DISABLE_MAPPING=true` on the unconfigured/passthrough proxy. Confirmed leak-free in testing, and correct for a deployment that does no encryption.

## Root cause

In passthrough, the backend short-circuits before per-statement cleanup:

- In `packages/cipherstash-proxy/src/postgresql/backend.rs`, the `if self.context.is_passthrough() { write; return }` branch returns **before** `complete_execution()` + `finish_session()`, which are the only drains for the per-connection `execute` and `session_metrics` queues.
- The frontend still enqueues for **every** statement (`start_session()` and `set_execute*()`).
- Net: each passthrough statement adds queue entries that are never popped → unbounded growth (~1 KB/statement measured).

Why only empty config: `is_passthrough() = encrypt_config.is_empty() || mapping_disabled()`. With a config present, even plaintext statements are `is_passthrough() = false` → cleanup runs → no leak.

Why `DISABLE_MAPPING` avoids it: that path makes the frontend early-return *before* `start_session()`, so the queues are never populated.

## Reproduction

Linux/glibc container, 6 rounds × 50k cached extended-protocol inserts (300k statements), identical load.

| Version / config | baseline | after 6 rounds | post-cooldown |
| -- | -- | -- | -- |
| 2.2.1, empty config | 36 MiB | 315 MiB | **317 MiB — no release** |
| 2.1.23, empty config | 28 MiB | 84 MiB | **83 MiB — no release** |
| 2.2.1, config present | 24 MiB | 37 MiB | 37 MiB (flat) |
| 2.2.1, empty config + `DISABLE_MAPPING=true` | 31 MiB | 34 MiB | 22 MiB (flat) |

Allocator is irrelevant (glibc == jemalloc); the leak is logical retention, not fragmentation.

## Proposed fix

In the passthrough branch, call `complete_execution()` + `finish_session()` before returning (or stop the frontend enqueuing execute/session state for passthrough statements). Add a regression test asserting the `execute` / `session_metrics` queues stay bounded under passthrough load.

coderdan:
Closes #400

## Summary

Fixes a memory leak (#400) that affects proxies running in **passthrough with an empty encrypt config** (no encryption configured). RSS climbs ~1 KB per statement until the pod is OOM-killed. Configured deployments are unaffected. Present in 2.1.23 and 2.2.1; reproduced locally and root-caused.

## Root cause

The frontend starts a session and enqueues an execute for **every** statement (`start_session` / `set_execute`). Those per-connection `execute` and `session_metrics` queues are only drained by `complete_execution()` / `finish_session()`. The passthrough branch in `backend.rs` returned **before** reaching those calls, so in passthrough the queues grew by one entry per statement and were never reclaimed.

Only manifests with an empty config because `is_passthrough() = encrypt_config.is_empty() || mapping_disabled()` — with a config present, even plaintext statements take the mapped path (which drains).

## Fix

Drain the `execute` and `session_metrics` queues on execute-terminating messages (`CommandComplete` / `EmptyQueryResponse` / `PortalSuspended` / `ErrorResponse`) in the passthrough branch too, mirroring the mapped path.

## Verification

Reproduced in a Linux/glibc container, 6×50k = 300k cached extended inserts, empty config:

| | before fix | after fix |
| --- | --- | --- |
| RSS, baseline → 300k inserts | 36 → **317 MiB (climbing, no release)** | 22 → **49 MiB (plateau)** |

Now matches the configured (non-leaking) case. Adds regression test `statement_lifecycle_does_not_grow_queues`. `cargo fmt` + `clippy --all-features -D warnings` clean.

## Follow-up

The memory reproduction harness (`scripts/mem-repro/`) and analysis write-up will land in a separate PR.

## Workaround (for customers on existing releases)

Set `CS_DEVELOPMENT__DISABLE_MAPPING=true` on unconfigured/passthrough proxies — confirmed leak-free.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented unbounded queue growth in the PostgreSQL proxy passthrough path, avoiding potential memory exhaustion when passthrough behavior is used.

* **Tests**
  * Added a regression test that repeatedly exercises passthrough execution and verifies execute/session queues are drained to prevent future leaks.

### DIFF
--- a/packages/cipherstash-proxy/src/postgresql/backend.rs
+++ b/packages/cipherstash-proxy/src/postgresql/backend.rs
@@ -726,3 +726,133 @@ where
         Ok(())
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::config::{LogConfig, TandemConfig};
+    use crate::log;
+    use crate::postgresql::context::KeysetIdentifier;
+    use crate::postgresql::messages::Name;
+    use crate::proxy::{EncryptConfig, EncryptionService};
+    use eql_mapper::Schema;
+    use std::io::Cursor;
+    use std::sync::Arc;
+    use tokio::sync::mpsc;
+
+    struct TestService {}
+
+    #[async_trait::async_trait]
+    impl EncryptionService for TestService {
+        async fn encrypt(
+            &self,
+            _keyset_id: Option<KeysetIdentifier>,
+            _plaintexts: Vec<Option<cipherstash_client::encryption::Plaintext>>,
+            _columns: &[Option<Column>],
+        ) -> Result<Vec<Option<crate::EqlCiphertext>>, Error> {
+            Ok(vec![])
+        }
+
+        async fn decrypt(
+            &self,
+            _keyset_id: Option<KeysetIdentifier>,
+            _ciphertexts: Vec<Option<crate::EqlCiphertext>>,
+        ) -> Result<Vec<Option<cipherstash_client::encryption::Plaintext>>, Error> {
+            Ok(vec![])
+        }
+    }
+
+    /// Builds a Context in passthrough mode (empty encrypt config), which is the
+    /// configuration that triggers BUG-300.
+    fn passthrough_context() -> Context<TestService> {
+        let config = Arc::new(TandemConfig::for_testing());
+        let encrypt_config = Arc::new(EncryptConfig::default());
+        let schema = Arc::new(Schema::new("public"));
+        let (reload_sender, _reload_receiver) = mpsc::unbounded_channel();
+
+        Context::new(
+            1,
+            config,
+            encrypt_config,
+            schema,
+            TestService {},
+            reload_sender,
+        )
+    }
+
+    /// Encodes a `CommandComplete` backend message on the wire:
+    /// `'C'` + Int32 length (body + 4) + null-terminated command tag.
+    fn command_complete_bytes() -> BytesMut {
+        let tag = b"SELECT 1\0";
+        let len = (tag.len() + 4) as i32;
+
+        let mut bytes = BytesMut::new();
+        bytes.extend_from_slice(b"C");
+        bytes.extend_from_slice(&len.to_be_bytes());
+        bytes.extend_from_slice(tag);
+        bytes
+    }
+
+    /// Regression test for BUG-300 (passthrough memory leak).
+    ///
+    /// The frontend enqueues a session + execute for *every* statement. Those
+    /// per-connection `execute` / `session_metrics` queues are only drained by
+    /// `complete_execution()` / `finish_session()`. Before the fix, the
+    /// passthrough branch in `rewrite()` returned early without calling these,
+    /// so the queues grew by one entry per statement and leaked until OOM.
+    ///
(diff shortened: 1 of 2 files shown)


packages/cipherstash-proxy/src/postgresql/context/mod.rs
```diff
@@ -1176,6 +1176,37 @@ mod tests {
...
 
+    /// Regression test for BUG-300 (passthrough memory leak).
+    ///
+    /// The frontend starts a session and enqueues an execute for every
+    /// statement — including passthrough statements. The `execute` and
+    /// `session_metrics` queues are only drained by `complete_execution()` /
+    /// `finish_session()`. Before the fix, the passthrough path in the backend
+    /// returned without calling these, so the queues grew by one entry per
+    /// statement and leaked memory until OOM. This asserts that a full
+    /// statement lifecycle leaves both queues empty, regardless of how many
```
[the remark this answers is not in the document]
coderdan: Valid.

packages/cipherstash-proxy/src/postgresql/backend.rs
```diff
@@ -705,3 +726,133 @@ where
...
+            encrypt_config,
+            schema,
+            TestService {},
+            reload_sender,
+        )
+    }
+
+    /// Encodes a `CommandComplete` backend message on the wire:
+    /// `'C'` + Int32 length (body + 4) + null-terminated command tag.
+    fn command_complete_bytes() -> BytesMut {
```
tobyhede: Might be worth expanding coverage for ALL message types

Could use a table-driven test that runs the same assertion for each terminating code:

```
  for msg_bytes in [
      command_complete_bytes(),
      error_response_bytes(),
      portal_suspended_bytes(),
      empty_query_response_bytes(),
  ] {
      // enqueue session + execute, run rewrite(), assert both queues == 0
  }
```

tobyhede (approved):

packages/cipherstash-proxy/src/postgresql/backend.rs
coderdan: Good call — done in 8f610ab3. Made it table-driven over all four execute-terminating codes (`CommandComplete`, `EmptyQueryResponse`, `PortalSuspended`, `ErrorResponse`), factoring a `backend_message()` wire-framing helper plus one encoder per code.

Confirmed it earns its keep: removing any single arm from the passthrough drain match now fails the test (e.g. dropping `PortalSuspended` → `PortalSuspended: execute queue not drained at statement 0`). Previously only `CommandComplete` was exercised, so three arms could regress silently.

PR #397, opened 2026-05-30
feat(cli): connection flags + DX (quiet logging, random port, best-effort TLS)

coderdan:
Implements CIP-3823, making the proxy usable as a plain CLI (for example, `npx stash proxy`) without hand-setting connection environment variables.

## Connection flags

- Add `--database-url <URL>` to parse a PostgreSQL connection string via `tokio-postgres`; individual `--db-*` flags take precedence.
- Add `--db-port` (`-P`) and `--db-password` (`-W`).
- Fix `--db-user` to set `CS_DATABASE__USERNAME` instead of the unused `CS_DATABASE__USER`.
- Reject non-UTF-8 URL passwords instead of silently retaining an absent or stale password.

## CLI behavior

- Log errors only by default while always printing the bound listen address.
- Preserve explicit `--log-level` and `CS_LOG__LEVEL` values, including an explicit `info`; an explicit level or config file takes precedence over `--debug`.
- Remove informational configuration and override chatter from quiet startup.
- If the implicit default listen port is occupied, bind an OS-assigned port and report it. Explicitly configured ports remain strict.
- Treat invalid inbound TLS as best effort: fall back to plaintext with a notice. `--tls` makes TLS mandatory, while `--no-tls` forces plaintext.
- Log PostgreSQL errors forwarded to clients concisely at debug level instead of dumping the full response at error level.

## Validation

- `cargo clippy --all-features -- -D warnings`
- `cargo check -p cipherstash-proxy-integration`
- `cargo test -p cipherstash-proxy database_url_sets_connection_env`
- `cargo test -p cipherstash-proxy bind_falls_back_to_random_port_when_default_in_use`
- `cargo test -p cipherstash-proxy log_level_preserves_whether_it_was_explicitly_set`
- Full `cipherstash-proxy` unit suite: 136/137 passed in parallel; the remaining environment-mutating config test passed in isolation.

The branch is rebased onto current `origin/main`, and all five commits include DCO sign-offs.

### DIFF
--- a/packages/cipherstash-proxy-integration/src/migrate/mod.rs
+++ b/packages/cipherstash-proxy-integration/src/migrate/mod.rs
@@ -42,9 +42,15 @@ mod tests {
             log_level: LogLevel::Debug,
             log_format: LogFormat::Pretty,
             command: None,
+            database_url: None,
             db_host: None,
+            db_port: None,
             db_name: None,
             db_user: None,
+            db_password: None,
+            no_tls: false,
+            tls: false,
+            debug: false,
         };
 
         let config = match TandemConfig::load(&args) {
--- a/packages/cipherstash-proxy-integration/src/migrate/mod.rs
+++ b/packages/cipherstash-proxy-integration/src/migrate/mod.rs
@@ -39,7 +39,7 @@ mod tests {
 
         let args = Args {
             config_file_path: "".to_string(),
-            log_level: LogLevel::Debug,
+            log_level: Some(LogLevel::Debug),
             log_format: LogFormat::Pretty,
             command: None,
             database_url: None,
--- a/packages/cipherstash-proxy/src/cli/mod.rs
+++ b/packages/cipherstash-proxy/src/cli/mod.rs
@@ -21,10 +21,22 @@ const DEFAULT_CONFIG_FILE: &str = "cipherstash-proxy.toml";
 /// CipherStash Proxy keeps your sensitive data in PostgreSQL encrypted and searchable, with no changes to SQL.
 ///
 pub struct Args {
+    /// Optional full PostgreSQL connection string, e.g.
+    /// "postgres://user:pass@host:5432/dbname".
+    /// Sets host, port, user, password and database name at once.
+    /// Individual flags below override matching parts of the URL.
+    #[arg(long, value_name = "URL", verbatim_doc_comment)]
+    pub database_url: Option<String>,
+
     /// Optional database host to connect to.
     /// Uses env or config file if not specified.
     #[arg(short = 'H', long)]
     pub db_host: Option<String>,
+
+    /// Optional database port to connect to.
+    /// Uses env or config file if not specified.
+    #[arg(short = 'P', long)]
+    pub db_port: Option<u16>,
 
     /// Optional database name to connect to.
     /// Uses env or config file if not specified.
@@ -35,6 +47,20 @@ pub struct Args {
     /// Uses env or config file if not specified.
     #[arg(short = 'u', long)]
     pub db_user: Option<String>,
+
+    /// Optional database password.
+    /// Uses env or config file if not specified.
+    /// Prefer CS_DATABASE__PASSWORD or --database-url to avoid leaking the
+    /// password into shell history / the process list.
+    #[arg(short = 'W', long, verbatim_doc_comment)]
+    pub db_password: Option<String>,
+
+    /// Disable inbound (client-facing) TLS: the proxy listens for client
+    /// connections in plaintext. Overrides any CS_TLS__* env / config.
+    /// Use for local development only. Does not affect the connection from the
+    /// proxy to the database.
+    #[arg(long, verbatim_doc_comment)]
+    pub no_tls: bool,
 
     /// Optional path to a CipherStash Proxy configuration file.
     ///
--- a/packages/cipherstash-proxy/src/cli/mod.rs
+++ b/packages/cipherstash-proxy/src/cli/mod.rs
@@ -59,8 +59,19 @@ pub struct Args {
     /// connections in plaintext. Overrides any CS_TLS__* env / config.
     /// Use for local development only. Does not affect the connection from the
     /// proxy to the database.
-    #[arg(long, verbatim_doc_comment)]
+    #[arg(long, verbatim_doc_comment, conflicts_with = "tls")]
     pub no_tls: bool,
+
+    /// Require inbound (client-facing) TLS. Startup fails if TLS is not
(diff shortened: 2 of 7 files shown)


coderdan:
## Review feedback addressed (pushed in c928bdfc)

Thanks for the reviews. Status:

**Fixed in this PR**
- **CI failure (the red Test jobs):** root cause was a compile error, not test logic — the new `Args` fields weren't added to the struct literal in `cipherstash-proxy-integration/src/migrate/mod.rs`, so the integration crate failed to build (`E0063`). Added the six missing fields.
- **Non-UTF8 password (Copilot + CodeRabbit, `tandem.rs`):** now returns a clear `InvalidParameter` error instead of silently dropping the password (which would connect with no/stale credentials). Went with Copilot's error-out approach over warn-and-continue, since the password is required downstream. Added a regression test (`database_url_rejects_non_utf8_password`) — the branch is reachable because tokio-postgres stores the URL password as raw bytes without UTF-8 validation.
- **Stale doc comment (Copilot + CodeRabbit, `tandem.rs`):** removed the leftover `extract_missing_field_and_key` text above `apply_database_url`.

**Deferred to follow-up issues** (real, but pre-existing / behavioural and out of scope for this CLI-feature PR)
- **IPv6 listen-address formatting** (CodeRabbit, `connect/mod.rs`): both the primary `to_socket_address()` and the new `:0` fallback lack IPv6 bracketing — a pre-existing limitation the fallback inherits. → #402
- **SIGHUP reload vs. TLS downgrade** (CodeRabbit, `main.rs`): a startup TLS downgrade can make `has_network_config_changed` report a spurious change on reload. Conservative today (refuses reload), worth fixing properly. → #403

Local validation: `cargo fmt --all --check`, `cargo clippy --all --all-features --tests -- -D warnings`, and the affected unit tests all pass.

tobyhede:
Reviewed at `fde3ad22`. Each item below is anchored to code I read on the branch head. Note that `fde3ad22` landed after CodeRabbit paused, so items 1 and 5 are in unreviewed code.

## Fix before merge

**1. `main.rs:30` — the quiet default discards the whole log config, not just the level.**

`LogConfig::with_level()` (`config/log.rs:99-111`) constructs from `Default` and copies nothing from the config that was just loaded. So when neither `CS_LOG__LEVEL` nor a config file is present, `format`, `output`, `ansi_enabled`, `slow_statements` and every per-target level are dropped. `CS_LOG__MAPPER_LEVEL=debug` — the recipe in `CLAUDE.md` and the logging docs — then produces no mapper output at all, and `CS_LOG__OUTPUT=stderr` still writes to stdout. `main` did `log::init(config.log.clone())`, so this is a regression for env-only deployments (the container ships no toml).

Suggest cloning `config.log` and overriding only `level` / `targets`.

**2. `main.rs:239` — an invalid certificate clears an explicitly configured `require_tls`.** 

The fallback arm sets `config.tls = None` *and* `config.server.require_tls = false`. `require_tls` is enforced at exactly one site (`postgresql/handler.rs:215`, via `context/mod.rs:880`), so clearing it removes the only check that rejects a client which skipped `SSLRequest`. An operator who set `CS_SERVER__REQUIRE_TLS=true` and then loses the cert (rotation, bad mount, unreadable file — all reach this arm via `config/tls.rs`) gets a proxy that starts, exits 0, and accepts plaintext sessions. On `main` the same input exited `exitcode::CONFIG`.

Two further notes: this arm emits only `eprintln!`, no tracing event, so nothing reaches structured logs or alerting; and `init()` also runs on the SIGHUP path (`main.rs:330`), so this can downgrade a proxy that is already serving traffic.

The plaintext fallback itself is a reasonable DX default. Discarding an explicit `require_tls = true` is the part that should stay fatal.

**3. `cli/mod.rs:65-69` — `--tls` does not require TLS.**

The help text reads "Require inbound (client-facing) TLS", but `args.tls` only reaches `init` as `force_tls`, where the valid-cert arm (`main.rs:224`) matches it as `_`. A repo-wide grep for `require_tls` shows three writes, all `false` (`config/server.rs:42`, `config/tandem.rs:108`, `main.rs:239`) — nothing ever sets it true. So `cipherstash-proxy --tls` with a valid certificate still serves a client that omits `SSLRequest`. `--no-tls` correctly writes the config at `tandem.rs:108`; `--tls` needs the symmetric write.

**4. `connect/mod.rs:67` — the random-port fallback pre-empts the retry loop.**

The fallback guard arm is matched before the retry arm at `:83` and returns on the first `AddrInUse`, so the backoff loop is unreachable for that error kind whenever fallback is on. `allow_random_fallback` is `!port_configured` (`main.rs:74`), which is true for any run with no `CS_SERVER__PORT` and no config file — including the reference `docker-compose.yml` service. If anything else already holds 6432 (double-start, stale process, unrelated service), the proxy binds a random port and reports success where `main` exited non-zero after ~1.5s of retries. The fallback port is printed, but exit code is 0 and nothing is emitted at `error!`/`warn!`.

Suggest falling back only once `retry_count > MAX_RETRY_COUNT`.

## Question

**5. `main.rs:33-39` — should file *existence* outrank `--debug`?**

`log_explicitly_set` includes `config_file_path.exists()` and is tested before `args.debug`, so `--debug` is a no-op whenever a `cipherstash-proxy.toml` is present, even when that file sets no `[log]` level at all. `cli/mod.rs:71-72` documents this, so I take it as intentional — but the comment at `main.rs:26` lists `--debug` first, and an explicit flag losing to the mere presence of a file is surprising. Worth confirming which behaviour you want, and aligning the two comments either way.

## Minor

- `postgresql/backend.rs:330-332` — the doc comment still says "All errors are logged at ERROR level for debugging and recorded in monitoring systems" while the body now logs at `debug!`. The demotion looks right (proxy-origin errors are already logged at `error!` in `frontend.rs`), but the comment is now false.
- `config/tandem.rs:391` — a Unix-socket URL yields no `Host::Tcp`, so `CS_DATABASE__HOST` is left alone while `CS_DATABASE__PORT` is still set from the same URL; multi-host URLs silently take the first. Leaving unspecified components to env/config is documented and matches libpq, so this is only about components the URL *does* contain.
- No CHANGELOG entries for the quiet-by-default level, the error-log demotion, or the random-port fallback — all three are operator-visible.

## Already tracked

`main.rs:322` — the reload guard runs before `init` at `:330`, so after a TLS fallback the running `tls: None` never matches a freshly loaded `tls: Some(..)` and every SIGHUP fails. This is CodeRabbit's comment, deferred to #403. One thing to add there: it is not just the cert-fix reload that fails — after a fallback, *all* reloads fail (log level, encrypt config) until the process restarts.

PR #405, opened 2026-06-01
ci: guard releases against version/changelog drift

coderdan:
## Summary

Stops releases from going out with a stale crate version or missing changelog entry — the drift that left **v2.2.1** shipping with `Cargo.toml` still at `2.2.0-alpha.1` and no `[2.2.1]` changelog section.

Two layers, plus docs:

### 1. CI guard (`.github/workflows/release.yml`)
New `verify-release` job that, on a numbered release, fails unless:
- the tag (minus `v`) equals the `[workspace.package]` version in `Cargo.toml`, and
- `CHANGELOG.md` contains a matching `## [X.Y.Z]` section.

The `build` matrix now `needs: verify-release`, so a mismatched tag **never builds or publishes**. On push/PR/`workflow_dispatch` the check is a no-op (gated by `if: github.event_name == 'release'`), so existing `main`-tag builds are unaffected.

### 2. `mise run release` (`mise.toml`)
Now verifies the same invariants *client-side* before tagging:
- tag matches `vX.Y.Z` (optional pre-release suffix),
- on a clean `main` that's in sync with `origin/main`,
- `Cargo.toml` version and `CHANGELOG.md` section already describe the version,
- tag doesn't already exist.

Previously it blindly tagged whatever was checked out — which is how the version bump got skipped.

### 3. Docs (`DEVELOPMENT.md`)
Documents the **prepare-release PR → tag** flow so the version bump + changelog entry are a reviewed step, not something remembered at tag time.

## Design note

The release stays **PR-prepared, not task-mutated**: the version bump and changelog land in a reviewable PR (e.g. #404 for v2.2.2), and `mise run release` only *gates and tags*. This fits branch protection on `main` and keeps the changelog honest between releases.

## Validation

- `mise.toml` parses (`tomllib`); extracted script passes `bash -n`; no-arg run fails fast; tag regex accepts `v2.2.2`/`v2.2.0-alpha.1`, rejects `2.2.2`/`v2.2`/`vfoo`.
- `release.yml` parses (`yaml.safe_load`); job graph is `verify-release → build → merge`.

Stacks logically after #404 (which adds the v2.2.2 version bump + changelog the guard checks for).

## Release Notes

* **Chores**
  * Added a release validation step that gates builds during releases and enforces matching release tag, workspace version, and corresponding changelog entry
  * Tightened the release task with pre-flight checks: version format, main-branch requirement, clean working tree, local/remote sync, and existing-tag detection

* **Documentation**
  * Updated releasing docs to a two-step process with version placeholders and clearer release instructions

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -16,8 +16,38 @@ env:
   REGISTRY_IMAGE: cipherstash/proxy
 
 jobs:
+  verify-release:
+    name: Verify release metadata
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+    steps:
+      - uses: actions/checkout@v4
+
+      # Only enforced for numbered releases. On push/PR/workflow_dispatch this
+      # job is a no-op so it can still gate the build matrix below.
+      - name: Check version + changelog match the release tag
+        if: github.event_name == 'release'
+        run: |
+          tag='${{ github.event.release.tag_name }}'
+          version="${tag#v}"
+
+          cargo_version="$(sed -n 's/^version = "\(.*\)"/\1/p' Cargo.toml | head -1)"
+          if [ "$cargo_version" != "$version" ]; then
+            echo "::error::Cargo.toml workspace version ($cargo_version) does not match release tag $tag. Bump the version in a prepare-release PR before tagging."
+            exit 1
+          fi
+
+          # Fixed-string match so dots in the version aren't treated as regex wildcards.
+          if ! grep -qF "## [$version]" CHANGELOG.md; then
+            echo "::error::CHANGELOG.md has no '## [$version]' section. Add release notes in a prepare-release PR before tagging."
+            exit 1
+          fi
+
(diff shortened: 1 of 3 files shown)


coderdan:
Rebased onto `main` (now includes #404) and addressed the Copilot review:

- **Regex wildcard in changelog match** (`mise.toml` + `release.yml`): switched `grep -q "^## \[$VER\]"` → `grep -qF "## [$VER]"` so dots in the version are literal. Verified `2.2.2` no longer false-matches `## [2x2x2]`.
- **`git rev-parse "$VERSION"` too permissive** (`mise.toml`): replaced with an explicit local check (`git rev-parse -q --verify "refs/tags/$VERSION"`) plus a remote check (`git ls-remote --exit-code --tags origin "refs/tags/$VERSION"`), so it only blocks on an actual existing tag and also catches remote-only tags.

Re-validated: `mise.toml` parses (`tomllib`), release task passes `bash -n`; `release.yml` parses and the job graph is still `verify-release → build → merge`.

tobyhede (approved):

PR #409, opened 2026-06-24
feat: client 0.38.0 migration — drop vendored stack-auth + EQL v2.3 mapper support (CIP-3233)

freshtonic:
## Summary

Moves Proxy off the `vendor/stack-auth` `[patch.crates-io]` workaround onto the released `cipherstash-client` group (CIP-3233 follow-up) — **and** carries the proxy-side work to keep the EQL/encrypted-query path working against the v2.3 payload format introduced by the 0.38.x client.

The 0.38.0 bump broke 38 integration tests (v2.3 changed how encrypted values are stored/searched). **All are fixed and CI is green across PostgreSQL 14–17** on the published **`cipherstash-client 0.38.1`**. Base branch: **`v3`**.

> [!NOTE]
> **Green and merge-ready** (pending review). Now on the published `0.38.1` (the stable release of the scalar-OPE `op`-store fix, CIP-3280 / suite #2061); the temporary CI-only prerelease pin was reverted and re-pointed at the real release.

## What's in this PR

| Area | Commits | What |
|---|---|---|
| **Deps / setup** (CIP-3233) | `2e11f69d`, `b38eeb0a`, `7b8500a2` | client `0.34.1-alpha.4 → 0.38.1` + drop vendored stack-auth/patch/exclude; EQL pin `→ eql-2.3.1`. |
| **jsonb range/order/projection** (CIP-3279) | `d1abdd2b` | `RewriteJsonbSteVecOrdering` → CLLW `ste_vec_entry` extractor form; proxy reshapes/decrypts bare sv entries. |
| **concrete-index-aware mapper + scalar OPE** (CIP-3283) | `1af99682`, `7248515a` | side-channel `IndexResolver` (no inference/unifier change); `RewriteScalarOpeOrdering` → `decode((col)::jsonb ->> 'op','hex')` bytea compare — built-ins only, **no EQL change**. |
| **jsonb equality** (CIP-3281) | `df28b0f8` | `=`/`<>` via XOR-aware `eql_v2.eq_term` (`hm`/`oc`); no EQL/param-encoding change. |
| **e2e-surfaced fixes** | `4f3c2a45`, `af5c963d` | `::jsonb` cast before `->> 'op'`; bare numeric sv literals encrypt as scalar `Plaintext::Float`. |
| **review feedback** | `76a8bfc5`, `fd20beab` | `SteVecTerm` fails fast w/o ste_vec index (CodeRabbit); dedup, module tests, perf (Toby). |
| **CI** | `ed7d29e3` | add `v3` to Test + Benchmark branch filters (PR retargeted to `v3`). |
| **deferred / housekeeping** | `1ad4c3f4`→`93ac7b1d`, `10b06cf5`→`717b960a` | temp prerelease pin (reverted); bool sv-leaf equality deferred (see below). |

## Why the extra scope

`cipherstash-client 0.38.x` (v2.3 EQL payload) changed encrypted storage/search: JSONB leaves became STE-vec elements (CLLW `oc` / HMAC `hm`), and scalar OPE columns no longer carry a root ordering term the bare operators read. CIP-3279/3281 rewrite jsonb comparisons to the `ste_vec_entry` extractor forms; CIP-3283 makes the mapper concrete-index-aware so scalar OPE ordering rewrites to a built-in `op` bytea compare — **all Proxy-side, no EQL change** (so **CIP-3282 / the EQL U-005 reversal is cancelled as unnecessary**).

## Verification

**CI green on PG 14–17** (`717b960a`, against published `0.38.1`): integration `without_multitenant` 348 + `multitenant` 69 + `showcase`; eql-mapper 122 unit; proxy lib 128; `mise run check` clean.

## Deferred (not in scope, tracked)

**jsonb sv boolean-leaf equality** (`col -> 'flag' = <bool>`) — **CIP-3293**. It was never implemented (only bool *containment* worked) and can't be wired Proxy-side: the SteVecTerm query encoding requires an *orderable* term, but a boolean has only the `hm` equality term — needs a `cipherstash-client` change. The proxy-side bool work was reverted; pre-existing scalar bool equality is unaffected.

## Review

All inline threads resolved — **9 CodeRabbit + 5 @tobyhede**. Real fixes from review: 2 (`SteVecTerm` fail-fast; the rest false-positives or quality/test/perf).

## Notes

- Integration tests need live Postgres + ZeroKMS (CI / creds).
- Tracking: CIP-3233 (umbrella) · CIP-3279 · CIP-3280 · CIP-3281 · CIP-3283 · CIP-3293 (deferred bool) · ~~CIP-3282~~ (cancelled, superseded by CIP-3283).


packages/cipherstash-proxy-integration/src/map_ope_index_order.rs
```diff
@@ -1,3 +1,14 @@
+// TODO(CIP-3280): add a NULL-invariant end-to-end test for the scalar OPE
```
tobyhede: Stale TODO?

packages/eql-mapper/src/ste_vec_ordering.rs
```diff
@@ -0,0 +1,144 @@
...
+        || matches!(parts.as_slice(), [schema, f] if schema == "eql_v2" && f == "jsonb_path_query_first")
+}
+
+/// Returns `true` if `expr` is EQL-typed in `node_types`.
+///
+/// Shared by the ordering and equality rewrite rules so both bind the rewrite
+/// to exactly the same operands. Kept here (rather than duplicated per rule)
+/// because mis-identifying an operand as EQL-typed would bind a rewrite to a
+/// value encrypted differently than the SQL extracts it.
+pub(crate) fn is_eql_typed(node_types: &HashMap<NodeKey<'_>, Type>, expr: &Expr) -> bool {
```
tobyhede, line 90: Duplicate implementations could be condensed and reused 

See `rewrite_scalar_ope_ordering.rs:119`

packages/cipherstash-proxy/src/postgresql/data/from_sql.rs
```diff
@@ -314,6 +396,37 @@ where
...
+/// encoding of the query term matches how the stored jsonb document's numeric
+/// leaves are encoded: cipherstash-client's STE-vec storage path always
+/// converts a JSON number leaf to its `f64` orderable representation (see
+/// `OrderableTerm::Number` derived via `f64::to_orderable_bytes`). Encoding the
+/// query term as an integer (`Plaintext::Int` / `BigInt`) would use a different
+/// orderable byte representation and produce incorrect comparison results.
+///
+/// Non-scalar values (objects, arrays) and JSON `null` are rejected — these
+/// term comparisons are only defined against scalar leaves.
+fn json_scalar_to_plaintext(value: &serde_json::Value) -> Result<Plaintext, MappingError> {
```
tobyhede, line 421: Claude reports:

> Bool leaf inconsistency (Rust M1): from_sql.rs:345-354 json_scalar_to_plaintext rejects JSON booleans with CouldNotParseParameter, but the equality rewrite docs + zerokms.rs:201 claim bool-leaf support, and Plaintext supports bools. Add a Value::Bool arm or tighten the docs.

packages/eql-mapper/src/ste_vec_ordering.rs
```diff
@@ -0,0 +1,144 @@
...
+//!   encrypted as the matching STE-vec query term (`oc` for CLLW ORE leaves,
+//!   `hm` for hmac/term-filter leaves).
+//!
+//! If these two passes disagree, the proxy can emit STE-vec SQL against a value
+//! that was *not* encrypted as a STE-vec query term (or vice versa), silently
+//! producing wrong results. Keeping the structural predicates **and** the
+//! encryption-binding-critical operand helpers ([`is_eql_typed`],
+//! [`rhs_as_jsonb`]) in one place ensures they cannot drift apart.
+
+use std::collections::HashMap;
```
tobyhede, line 23: Module lacks unit tests

packages/eql-mapper/src/eql_mapper.rs
```diff
@@ -239,8 +265,81 @@ impl<'ast> EqlMapper<'ast> {
...
+    /// (`->` / `->>` / `jsonb_path_query_first`).
+    ///
+    /// The returned sets identify the RHS values (params by [`Param`], literals
+    /// by [`NodeKey`]) that must be encrypted as a STE-vec query term
+    /// ([`EqlTerm::SteVecTerm`]) rather than a full/partial root payload.
+    /// Ordering binds the term to `eql_v2.ore_cllw(...)` (`oc`); equality binds
+    /// it to `eql_v2.eq_term(...)` (the XOR-aware `hm`/`oc` term the column's
+    /// leaf carries). Both require the same `SteVecTerm` reclassification — the
+    /// proxy's encrypt path emits whichever term the column's leaf carries.
+    fn ste_vec_term_rhs_keys(
```
tobyhede, line 293: [Claude]

> 
> Perf nit (Low) — ste_vec_term_rhs_keys runs twice per statement.
> 
> resolve() calls param_types() (uses params, drops literals) and literal_types() (uses literals, drops params); each independently calls ste_vec_term_rhs_keys (eql_mapper.rs:244 and :341). That's two full O(nodes) AST scans + per-operand peek_node_type/follow_tvars, producing identical results — half discarded each time.
> 
> Fix: compute it once in resolve() (~line 177) and pass each half into param_types/literal_types as an argument. Mechanical; no behavior change.
>

tobyhede:
## CodeRabbit review notes (verified)

Ran CodeRabbit + manually verified each finding against current code. Notes below cover the findings **in this PR's diff**. (Several higher-severity CodeRabbit findings were on *untracked* working-tree files — `bloom_credit/*.json` PII, `.github/secrets.env.plaintext`, `generate_ore_fixtures.rs`, `large_json_insert.rs`, `.serena/` — those are **not part of this PR** and are being handled separately.)

### Should address

- **`from_sql.rs` (~L349-374) — sv-term RHS as TEXT/VARCHAR mis-encrypted.** `SteVecTerm` + `ColumnType::Json` only matches the `JSON|JSONB|BYTEA` arm. A binary `TEXT`/`VARCHAR`-typed RHS falls through to the generic `(_, _, TEXT|VARCHAR)` arm, which calls `text_from_sql(val, EqlTermVariant::Full, …)` (hardcoded `Full`) and returns a full `Plaintext::Json` payload instead of the scalar sv-term. Add an explicit `(SteVecTerm, Json, TEXT|VARCHAR)` arm that preserves the `SteVecTerm` variant. Narrow trigger, but silently wrong when hit.

- **`eql_mapper.rs` (L244-256) — reused bind param reclassified globally.** `reclassify_as_ste_vec_term_if(value, ste_vec_params.contains(&p))` flips a param for *all* its occurrences. If the same `$n` is used in both a jsonb sv-term comparison and an ordinary encrypted predicate, it gets the wrong encrypted representation everywhere. Either error on mixed-use params or track typing per occurrence.

- **`manager.rs` (L105-119) — `InvalidEncryptionConfig` fails open.** On a *malformed* (not missing) config at startup we log and fall through to an empty `EncryptConfig::new()`, i.e. start serving with encryption disabled. Consider failing closed (return the error) for the invalid case, distinct from the missing-table bootstrap path.

### Known limitation (confirm intent)

- **`rewrite_jsonb_ste_vec_equality.rs` (L68-78) — reversed operand orientation.** `value = col -> selector` is intentionally left to fall back to root `eql_v2_encrypted` equality; the reclassification pass gates on the same `is_ste_vec_accessor(left)` predicate so the two passes stay in lockstep. Verified consistent and documented. Worth confirming the reversed form doesn't silently mis-compare (vs erroring) — if it can, a guard/error would be safer than a wrong result.

### Minor

- **`common.rs` (L94-96)** — `format!("TRUNCATE {}", table)` splices an identifier; test-controlled names only, but quoting the identifier removes the foot-gun.
- **`error.rs` (L188)** — `InvalidEncryptionConfig` has no docs link (some neighbors do); add one for the customer-facing path.
- **`CHANGELOG.md`** — current-stack entries are split across `Unreleased`/older sections; consolidate under 2.2.4.

tobyhede (approved):

Cargo.toml
```diff
@@ -45,9 +43,9 @@ debug = true
 
 [workspace.dependencies]
 sqltk = { version = "0.10.0" }
-cipherstash-client = { version = "=0.34.1-alpha.4" }
-cipherstash-config = { version = "=0.34.1-alpha.4" }
-cts-common = { version = "=0.34.1-alpha.4" }
+cipherstash-client = { version = "0.38.0" }
+cipherstash-config = { version = "0.38.0" }
+cts-common = { version = "0.38.0" }
```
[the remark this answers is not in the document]
freshtonic: All three crates are published at 0.38.0 on crates.io — `cipherstash-client-0.38.0.crate`, `cipherstash-config-0.38.0.crate` and `cts-common-0.38.0.crate` each return HTTP 200 from static.crates.io, and `Cargo.lock` pins them from the `registry+https://github.com/rust-lang/crates.io-index` source with checksums. The build resolves from the registry and CI is green on PG14–17, so this is a false positive.

packages/cipherstash-proxy/src/proxy/encrypt_config/index_resolver.rs
```diff
@@ -0,0 +1,188 @@
...
+}
+
+impl IndexResolver for EncryptConfigIndexResolver {
+    fn resolve(&self, table_column: &TableColumn) -> HashSet<IndexKind> {
+        // Mirror how the column mapper derives an Identifier from a TableColumn
+        // (unquoted ident text) so the lookup keys match.
+        let identifier = Identifier::new(
+            table_column.table.value.to_string(),
+            table_column.column.value.to_string(),
+        );
```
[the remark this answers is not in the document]
freshtonic: The resolver intentionally mirrors the existing `column_mapper.rs` lookup, which builds the same key via `Identifier::new(table.value.to_string(), column.value.to_string())` (column_mapper.rs:43/78/107). Both paths use bare unquoted names, so the resolver is consistent with established behaviour. Schema-qualified table keys are a pre-existing limitation across the whole EncryptConfig keying — not something this PR introduces or worsens — so there's no divergence to fix here.

packages/cipherstash-proxy/src/proxy/zerokms/zerokms.rs
```diff
@@ -201,6 +201,17 @@ impl EncryptionService for ZeroKms {
...
+                    // SteVecTerm generates a CLLW ORE comparison term (`oc`) for the
+                    // right-hand side of a jsonb sv ordering comparison
+                    // (`col -> selector <op> $param`).
+                    EqlTermVariant::SteVecTerm => col
+                        .config
+                        .indexes
+                        .iter()
+                        .find(|i| matches!(i.index_type, IndexType::SteVec { .. }))
+                        .map(|index| EqlOperation::Query(&index.index_type, QueryOp::SteVecTerm))
+                        .unwrap_or(EqlOperation::Store),
```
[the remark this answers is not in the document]
freshtonic: Good catch — confirmed reachable and **fixed** in 76a8bfc5. Every `eql_v2_encrypted` column is granted `JsonLike` in the eql-mapper schema (schema/manager.rs:146) irrespective of its configured indexes, and the config only enforces ste_vec ⟹ jsonb — so a `SteVecTerm` value can legitimately reach a column with no ste_vec index. The `SteVecTerm` arm now fails fast with `EncryptError::UnknownIndexTerm` instead of silently storing a payload that lacks the queried term; covered by three new unit tests in zerokms.rs.

packages/eql-mapper/src/transformation_rules/rewrite_jsonb_ste_vec_ordering.rs
```diff
@@ -0,0 +1,203 @@
...
+        Ok(false)
+    }
+
+    fn would_edit<N: Visitable>(&mut self, node_path: &NodePath<'ast>, _target_node: &N) -> bool {
+        if let Some((Expr::BinaryOp { left, op, right },)) = node_path.last_1_as::<Expr>() {
+            if is_ordering_operator(op)
+                && is_ste_vec_accessor(left)
+                && self.is_eql_typed(left)
+                && self.is_eql_typed(right)
+            {
```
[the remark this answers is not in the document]
freshtonic: This shape isn't representable. EQL terms are column-scoped, so `accessor <op> other_encrypted -> sel` (or any different encrypted column on the RHS) fails type-checking with "cannot unify EQL terms …" before any rewrite runs — verified by the new `jsonb_sv_ordering_rhs_different_encrypted_column_is_rejected` test. The only EQL RHS that survives unification is one bound to the *same* TableColumn (a literal/param, or the same column re-referenced), which carries the correct sv term, so the `is_eql_typed(right)` gate is sufficient.

packages/eql-mapper/src/transformation_rules/rewrite_scalar_ope_ordering.rs
```diff
@@ -0,0 +1,232 @@
...
+        if let Some((Expr::BinaryOp { left, op, right },)) = node_path.last_1_as::<Expr>() {
+            return matches!(
+                op,
+                BinaryOperator::Lt
+                    | BinaryOperator::LtEq
+                    | BinaryOperator::Gt
+                    | BinaryOperator::GtEq
+            ) && self.is_scalar_ope_column(left)
+                && self.is_eql_typed(right);
+        }
```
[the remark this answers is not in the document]
freshtonic: Same reasoning as the jsonb ordering rule: a different encrypted column on the RHS fails EQL term unification (`a > b` → "cannot unify EQL terms …"), verified by the new `scalar_ordering_rhs_different_encrypted_column_is_rejected` test. A param/literal RHS unifies to the *same* TableColumn and is therefore OPE-backed with its own `op` term, so `is_eql_typed(right)` cannot bind a non-OPE foreign column. No change needed.

packages/eql-mapper/src/type_checked_statement.rs
```diff
@@ -153,6 +164,11 @@ impl<'ast> TypeCheckedStatement<'ast> {
         DryRunnable::new((
             RewriteStandardSqlFnsOnEqlTypes::new(Arc::clone(&self.node_types)),
             RewriteContainmentOps::new(Arc::clone(&self.node_types)),
+            RewriteJsonbSteVecOrdering::new(Arc::clone(&self.node_types)),
+            RewriteScalarOpeOrdering::new(
+                Arc::clone(&self.node_types),
+                Arc::clone(&self.index_resolver),
```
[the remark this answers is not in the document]
freshtonic: The reclassification pass (eql_mapper.rs `literal_types`/`ste_vec_term_rhs_keys`) deliberately doesn't consult the IndexResolver — it gates on `is_ste_vec_accessor` + EQL-typedness, and ste_vec_ordering.rs documents that the rewrite and reclassification passes must agree. Gating only the rewrite on `resolve(tc).contains(SteVec)` would desync the two passes and reintroduce the mismatch. The actual unsafe outcome (a jsonb column without ste_vec) is now caught at the encryption layer by the SteVecTerm fail-fast (3473973385), so this gate is redundant and the naive form would be harmful.

packages/eql-mapper/src/ste_vec_ordering.rs
```diff
@@ -0,0 +1,144 @@
...
+    }
+
+    // Defensive fallback: cast whatever we have to JSONB.
+    Expr::Cast {
+        kind: CastKind::DoubleColon,
+        expr: Box::new(expr),
+        data_type: DataType::JSONB,
+        format: None,
+    }
+}
```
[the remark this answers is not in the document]
freshtonic: The transform is bottom-up — sqltk's `apply_transform` applies depth-first to children before the parent — so `CastLiteralsAsEncrypted`/`CastParamsAsEncrypted` cast the RHS child to `::JSONB::eql_v2_encrypted` *before* the ordering/equality rule sees the parent comparison. `rhs_as_jsonb` then strips that outer cast; the emitted SQL (`eql_v2.ore_cllw($2::JSONB)`, and `'ENC(4)'::JSONB` for literals) proves the strip branch is reached. It's live, load-bearing code, not dead, so I've left it as-is.

packages/eql-mapper/src/transformation_rules/rewrite_jsonb_ste_vec_equality.rs
```diff
@@ -0,0 +1,239 @@
...
+    /// carries.
+    fn build_eq_term_rhs(expr: Expr) -> Expr {
+        let rhs = rhs_as_jsonb(expr);
+
+        let hm = Self::field_text(rhs.clone(), "hm");
+        let oc = Self::field_text(rhs, "oc");
+        let coalesce = Self::call("", "coalesce", vec![hm, oc]);
+
+        Self::call("", "decode", vec![coalesce, Self::string_literal("hex")])
+    }
```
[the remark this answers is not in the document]
freshtonic: No leak. `CastLiteralsAsEncrypted` keys on the original AST literal node through `node_path`, and because the walk is bottom-up it casts that child to `::JSONB::eql_v2_encrypted` before the equality rule runs on the parent. `build_eq_term_rhs` therefore clones the already-encrypted node, not the plaintext, and the cast rule's "all literals consumed" postcondition still passes. I added `jsonb_sv_equality_rhs_literal_is_encrypted_in_sql`, which transforms the literal form and asserts the RHS appears only as `'ENC(4)'::JSONB ->> 'hm'/'oc'` with no bare `4` — i.e. the plaintext never survives.

packages/cipherstash-proxy/src/postgresql/data/from_sql.rs
```diff
@@ -417,6 +536,136 @@ mod tests {
...
+        log::init(LogConfig::default());
+
+        let literal = Value::Number(BigDecimal::from_str("-70").unwrap(), false);
+
+        let pt = literal_from_sql(&literal, EqlTermVariant::SteVecTerm, ColumnType::Json)
+            .unwrap()
+            .unwrap();
+
+        assert_eq!(pt, Plaintext::Float(Some(-70.0)));
+    }
```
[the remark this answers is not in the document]
freshtonic: The mechanism is slightly different from the description, and the outcome is safe rather than wrong. `-70` parses as `Expr::UnaryOp{Minus, Number}`, which `infer_type_impls/expr.rs` forces to `Native`, so `encrypted_jsonb -> 'number' > -70` fails type-checking outright ("cannot unify values Native and EQL…") — verified by two new tests. Since the UnaryOp RHS is Native, `is_eql_typed(right)` is false and the rewrite never fires, so there's no "SQL rewritten but value not reclassified" mismatch; the statement is cleanly rejected. This UnaryOp→Native behaviour is pre-existing (unchanged by this PR) and applies identically to scalar EQL (`encrypted_int > -70`), so a real fix belongs in core type inference, not the reclassification loop.

packages/cipherstash-proxy-integration/src/map_ope_index_order.rs
freshtonic: Rewritten. The NULL `ORDER BY` placement half is already covered by `map_ope_order_nulls_last_by_default`/`map_ope_order_nulls_first`, so I narrowed the TODO to the remaining gap (comparison against a NULL operand / three-valued logic) and corrected the blocker framing to "gated on bumping the client pin off 0.38.0" — the `op` slot is validated e2e on 0.38.1-alpha.1.

packages/eql-mapper/src/ste_vec_ordering.rs
freshtonic: Done. Removed the private `is_eql_typed` from both `rewrite_scalar_ope_ordering.rs` and `rewrite_containment_ops.rs` and pointed them at the shared `ste_vec_ordering::is_eql_typed`. The bodies were byte-identical, so no emitted-SQL change; all rewrite tests stay green.

packages/cipherstash-proxy/src/postgresql/data/from_sql.rs
freshtonic: Fixed — added the `Value::Bool` arm to `json_scalar_to_plaintext`, so a boolean sv leaf reduces to a scalar `Plaintext::Boolean` and is encrypted as the `hm` equality term (`col -> 'flag' = true` now works instead of `CouldNotParseParameter`). Verified unit-first (TDD) plus a self-contained `select_where_jsonb_bool_eq` integration test for CI; confirmed against the client that boolean leaves are equality-only (no `oc`/ordering term), so this can't open an ordering path. Validated unit-only locally (no live PG/ZeroKMS in my env) — the integration test runs in CI.

packages/eql-mapper/src/ste_vec_ordering.rs
freshtonic: Added 13 unit tests covering all the predicates/operand helpers (positive and negative), including the `eql_v2.`-qualified `jsonb_path_query_first` accessor form and the `rhs_as_jsonb` strip / defensive-wrap / no-peel cases. I mutation-checked the encryption-binding `rhs_as_jsonb` test to confirm it catches a regression.

packages/eql-mapper/src/eql_mapper.rs
freshtonic: Refactored — `ste_vec_term_rhs_keys` is now computed once in `resolve()` and its param/literal halves are threaded into `param_types`/`literal_types`, eliminating the second full AST scan. Both happy and error paths share the precomputed sets; no behaviour change, all tests green.

packages/cipherstash-proxy/src/postgresql/data/from_sql.rs
freshtonic: Follow-up / correction: the `Value::Bool` arm was necessary but **not sufficient** — bool sv-leaf equality fails end-to-end. e2e CI surfaced `E58000: Cannot convert value of type Option<bool> to orderable term`: the equality rewrite reclassifies the RHS to `SteVecTerm`, and cipherstash-client's `SteVecTerm` query encoding requires an *orderable* term, but a boolean leaf only has the `hm` term (no `oc`). It's also a never-implemented path (only bool *containment* worked before), not a regression. So per the "or tighten the docs" side of this comment, I've reverted the proxy-side bool work (`717b960a`) — the `Value::Bool` arm + its tests and the `select_where_jsonb_bool_eq` tests — and filed CIP-3293 for the cipherstash-client change needed to emit the `hm` term for non-orderable sv leaves. Pre-existing scalar bool equality is unaffected.

packages/eql-mapper/src/transformation_rules/rewrite_jsonb_ste_vec_ordering.rs
```diff
@@ -0,0 +1,153 @@
...
+///
+/// ```sql
+/// (col -> selector) > $param
+/// ```
+///
+/// into:
+///
+/// ```sql
+/// eql_v2.ore_cllw(col -> selector) > eql_v2.ore_cllw($param::jsonb)
+/// ```
```
coderdan, line 51: This may be addressed in EQL V3 but this approach locks Proxy in to CLWW. Probably not worth addressing for 2.3 but leaving as a comment for anyone who looks at this in the future (including Claude).

packages/eql-mapper/src/lib.rs
```diff
@@ -2088,6 +2091,711 @@ mod test {
         );
     }
 
+    /// Group B (CIP-3279): ordering comparisons on a jsonb sv-element extracted
```
coderdan, line 2094: Should only reference public GH issues in public repos.

packages/cipherstash-proxy-integration/src/map_ope_index_order.rs
```diff
@@ -1,3 +1,20 @@
+// TODO(CIP-3280): add an end-to-end test for `<`/`<=`/`>`/`>=` comparisons
```
coderdan, line 1: We should address the TODO. Also reference a public GH issue instead of Linear.

coderdan (approved):
Looks great! Just minor things that should be addressed.

PR #419, opened 2026-07-21
fix: satisfy clippy::for_kv_map in type_checked_statement

freshtonic:
CI on #418 surfaced a pre-existing `clippy::for_kv_map` failure — unrelated to that PR's lockfile deletion, but the deletion let the clippy step run to completion and expose it. CI runs clippy with `-D warnings`, so the lint is a hard error.

`packages/eql-mapper/src/type_checked_statement.rs:126` iterated a map with `.iter()` and immediately discarded the value:

```rust
for (key, _) in encrypted_literals.iter() {
```

Switched to `.keys()`:

```rust
for key in encrypted_literals.keys() {
```

Behaviour is identical; the value was never used.

## Verification

`RUSTFLAGS="-D warnings" cargo clippy --all --all-targets` is clean across the whole workspace (checked there was no second error hiding behind the one that stopped CI's build).

### DIFF
--- a/packages/eql-mapper/src/type_checked_statement.rs
+++ b/packages/eql-mapper/src/type_checked_statement.rs
@@ -123,7 +123,7 @@ impl<'ast> TypeCheckedStatement<'ast> {
             )));
         }
 
-        for (key, _) in encrypted_literals.iter() {
+        for key in encrypted_literals.keys() {
             if !self.literal_exists_for_node_key(*key) {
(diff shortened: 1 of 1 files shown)


freshtonic:
Superseded by #418, which now bundles this exact clippy fix alongside the stray-lockfile deletion so that PR is self-contained and green. Closing as redundant.

PR #423, opened 2026-07-21
feat: migrate to EQL v3 — cipherstash-client 0.42, typed jsonb domains, v3 typecheck & transform

freshtonic:
### 📚 eql-v3 PR &nbsp;·&nbsp; 1 of 1

✅🟢 **[#423 `queue/eql-v3/upgrade-deps`](https://github.com/cipherstash/proxy/pull/423) → `main`** &nbsp;👈 **this PR**

# About this queue

# About this branch

feat: migrate to EQL v3 — cipherstash-client 0.42, typed jsonb domains, v3 typecheck & transform


tobyhede:
## Review — v3 representation adoption

Solid, well-commented migration. One blocking item; rest are minor.

### 🔴 Blocking — all four `to_ciphertext_*` tests fail

`data_row.rs` `TryFrom<&mut DataColumn>` now only strips a leading `0x01`; everything else goes straight to `serde_json::from_slice`. But none of the four `to_ciphertext_*` fixtures were migrated off v2 wire format:

- text tests — bytes start with `("…")` composite wrapper (`0x28`) → parse error → `None`
- binary tests — bytes start with the v2 12-byte rowtype header (`0x00`), *not* `0x01` → parse error → `None`
- `_and_null` binary test also loses the v2 `ColumnIsNull` sentinel, so `set_null()` never fires

`as_ciphertext` swallows the errors to `None`, so every `assert!(encrypted[..].is_some())` fails. `cargo test -p cipherstash-proxy` → 4 red. The parser is correct for real v3 payloads — only the fixtures are stale. Regenerate all four as real v3 jsonb bytes (binary: `0x01` + JSON; text: bare `{…}`).

⚠️ This base branch runs no test CI (`gh pr checks` shows only CodeRabbit skipped), so this won't be caught automatically.

### 🟡 Minor

- **Dead code:** `[patch.crates-io] stack-auth` is removed but `vendor/stack-auth/` is still in the tree. Delete it so the "unreferenced" comment holds.
- **Version:** `mise.toml` / `mise.local.example.toml` pin `eql-3.0.1`, but the queue description targets `3.0.2`. Intentional?
- **Diagnostics:** `CouldNotDecryptDataForKeyset { keyset_id, .. }` drops the new `#[source]` zerokms error; the chain dies at Proxy's boundary. Acknowledged in-comment — fine to defer.
- **Error text:** `SteVecSelectorInvalid` interpolates the selector (tokenized index term) into its message → ends up in logs. Low risk.

### ℹ️ Unverified — worth a check against client 0.42.0

- `zerokms.rs` decrypt assumes `sv[0]` is the SteVec root; nothing in Proxy enforces root-first ordering — relies on the client preserving it.
- `Plaintext::from_slice` now carries the full burden of recovering the typed variant (Int/Text/Float…) that v2's `decrypt_eql` returned directly. Confirm the self-describing decode round-trips every type.

👍 Nice: `bytes.first()` match replaces the old `&bytes[0..=1]` that panicked on <2-byte columns.

packages/cipherstash-proxy/src/error.rs
```diff
@@ -254,6 +254,23 @@ pub enum EncryptError {
     #[error("InvalidIndexTerm")]
     InvalidIndexTerm,
 
+    /// EQL v3 orders encrypted jsonb entries by the CLLW-OPE (`op`) term and has
+    /// no representation for CLLW-ORE (`oc`), so a column configured for
+    /// Standard-mode ste_vec cannot be encrypted. The column has to be
+    /// reconfigured and its data re-encrypted.
+    #[error("An encrypted jsonb column is configured for ORE ordering, which EQL v3 does not support. For help visit {}#encrypt-ste-vec-ore-mode-unsupported", ERROR_DOC_BASE_URL)]
```
tobyhede, line 261: Entry needs to be added to `errors.md`

packages/cipherstash-proxy/src/error.rs
```diff
@@ -254,6 +254,23 @@ pub enum EncryptError {
...
+    SteVecOreModeUnsupported,
+
+    /// `sv[0]` is the decryption root of a SteVec document, so an empty `sv`
+    /// array leaves nothing to decrypt.
+    #[error("Encrypted jsonb value has no root entry and cannot be decrypted")]
```
tobyhede: Does this need to reference `errors.md`?

packages/cipherstash-proxy/src/error.rs
```diff
@@ -254,6 +254,23 @@ pub enum EncryptError {
...
+    SteVecMissingRootEntry,
+
+    /// A SteVec entry's selector is the source of both AEAD bindings (nonce and
+    /// AAD), so it must be exactly 16 hex-encoded bytes.
+    #[error("Encrypted jsonb entry has an invalid selector '{selector}'")]
```
tobyhede: Does this need to reference `errors.md`?

tobyhede (approved):

freshtonic:
Addressed the review (all pushed):

- **🔴 Blocking — the four `to_ciphertext_*` tests**: confirmed they fail. The fixtures are stale v2 payloads, and a valid v3 `EqlCiphertextV3` needs a real MessagePack-Base85 `c` ciphertext (and a real SteVec document for the jsonb case) produced by the live encrypt path — not something that can be hand-crafted correctly. Leaving these to be regenerated against a live ZeroKMS/DB rather than fabricating ciphertext bytes. Flagging explicitly so it isn't lost.
- **`errors.md`**: added the `#encrypt-ste-vec-ore-mode-unsupported` entry (it was a dangling link), and added `For help visit …` doc links + `errors.md` entries for `SteVecMissingRootEntry` and `SteVecSelectorInvalid` (see the inline-thread replies). → `docs(errors): document the EQL v3 SteVec errors`
- **Dead code**: removed `vendor/stack-auth/` and dropped the now-pointless workspace `exclude`. → `chore: remove the unreferenced vendored stack-auth`
- **Version**: bumped EQL `3.0.1 → 3.0.2` in `mise.toml` / `mise.local.example.toml` to match the queue target. → `chore(deps): pin EQL to 3.0.2`
- **Diagnostics / error-text minors** (`CouldNotDecryptDataForKeyset` dropping `#[source]`; `SteVecSelectorInvalid` interpolating the term): left as acknowledged/deferred, per your notes.

The `sv[0]` root-ordering and `Plaintext::from_slice` round-trip "unverified" items need a live v3 DB to exercise, so they remain open.

packages/cipherstash-proxy/src/error.rs
freshtonic: Done — added an **Encrypted jsonb column configured for ORE ordering** section under Encrypt errors with the `encrypt-ste-vec-ore-mode-unsupported` anchor (it was referenced here but the anchor didn't exist, so this was a dangling link).

packages/cipherstash-proxy/src/error.rs
freshtonic: Yes — added `For help visit …#encrypt-ste-vec-missing-root-entry` to the error and a matching **Encrypted jsonb value has no root entry** section under Decrypt errors.

packages/cipherstash-proxy/src/error.rs
freshtonic: Yes — added `For help visit …#encrypt-ste-vec-selector-invalid` to the error and a matching **Encrypted jsonb entry has an invalid selector** section under Decrypt errors.

PR #422, opened 2026-07-21
[1/7] docs: set up agent-skill conventions and domain glossaries

freshtonic:
### 📚 eql-v3 PR &nbsp;·&nbsp; 1 of 7

Part of a queue. The PRs merge in FIFO order — the numbered order below, #1 first. Merging one supersedes the PRs after it until the author runs `git queue sync` (rebases the rest onto the merged base) and `git queue submit` (retargets their PRs).

✅🟢 **[#422 `queue/eql-v3/setup-skills`](https://github.com/cipherstash/proxy/pull/422) → `main`** &nbsp;👈 **this PR**
✅🟢 [#423 `queue/eql-v3/upgrade-deps`](https://github.com/cipherstash/proxy/pull/423) → `queue/eql-v3/setup-skills`
♻️🟢 [#424 `queue/eql-v3/typecheck`](https://github.com/cipherstash/proxy/pull/424) → `queue/eql-v3/upgrade-deps`
✅🟢 [#428 `queue/eql-v3/transform`](https://github.com/cipherstash/proxy/pull/428) → `queue/eql-v3/typecheck`
✅🟢 [#426 `queue/eql-v3/showcase`](https://github.com/cipherstash/proxy/pull/426) → `queue/eql-v3/transform`
⏳🟢 [#427 `queue/eql-v3/integration`](https://github.com/cipherstash/proxy/pull/427) → `queue/eql-v3/showcase`
⏳🟢 [#430 `feat/eql-v3-jsonb-composite-selectors`](https://github.com/cipherstash/proxy/pull/430) → `queue/eql-v3/integration`

<sub>✅ approved · ♻️ changes requested · ⏳ review pending &nbsp;|&nbsp; 🟣 merged · 🟢 open · ⚫ closed &nbsp;—&nbsp; status as of the last `git queue submit`.</sub>
<sub>🥞 Managed by git-queue — do not edit this list by hand.</sub>

# About this queue

Migrates CipherStash Proxy from EQL v2 to EQL v3 (cipherstash-client 0.34.1-alpha.4 → 0.42.0, EQL 2.3.0-pre.3 → 3.0.2), replacing the opaque eql_v2_encrypted composite type with 53 typed jsonb domains that encode both scalar type and searchable capability in the column type itself.

# About this branch

Set up the Matt Pocock skills which will we use to drive the subsequent changes.

## Summary by CodeRabbit

* **Documentation**
  * Added guidance for tracking, triaging, and linking engineering issues.
[...]

PR #424, opened 2026-07-21
[2/6] feat: EQL v3 type checker driven by per-column domain capability

freshtonic:
### 📚 eql-v3 PR &nbsp;·&nbsp; 2 of 6

✅🟢 [#423 `queue/eql-v3/upgrade-deps`](https://github.com/cipherstash/proxy/pull/423) → `main`
♻️🟢 **[#424 `queue/eql-v3/typecheck`](https://github.com/cipherstash/proxy/pull/424) → `queue/eql-v3/upgrade-deps`** &nbsp;👈 **this PR**
✅🟢 [#428 `queue/eql-v3/transform`](https://github.com/cipherstash/proxy/pull/428) → `queue/eql-v3/typecheck`
✅🟢 [#426 `queue/eql-v3/showcase`](https://github.com/cipherstash/proxy/pull/426) → `queue/eql-v3/transform`
⏳🟢 [#427 `queue/eql-v3/integration`](https://github.com/cipherstash/proxy/pull/427) → `queue/eql-v3/showcase`
⏳🟢 [#430 `feat/eql-v3-jsonb-composite-selectors`](https://github.com/cipherstash/proxy/pull/430) → `queue/eql-v3/integration`

# About this queue

# About this branch

Type checker for v3 domains — extends EQL Mapper to emit `eql_v3_<token>_<capability>` domains instead of eql_v2_encrypted
(designed, not yet built; the step that makes the migration actually work).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -832,6 +832,7 @@ dependencies = [
  "clap",
  "config",
  "cts-common",
+ "eql-bindings",
  "eql-mapper",
  "exitcode",
  "hex",
@@ -1483,6 +1484,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "92773504d58c093f6de2459af4af33faa518c13451eb8f2b5698ed3d36e7c813"
 
 [[package]]
+name = "dyn-clone"
+version = "1.0.20"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "d0881ea181b1df73ff77ffaaf9c7544ecc11e82fba9b5f27b262a3c73a332555"
+
+[[package]]
 name = "either"
 version = "1.15.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1510,6 +1517,18 @@ dependencies = [
  "proc-macro2",
  "quote",
  "syn 2.0.117",
+]
+
+[[package]]
+name = "eql-bindings"
+version = "3.0.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "39a6e26dbefb089908aa10521c6efffbe3f5404f9adfa7d43e4244d41c1d6db0"
+dependencies = [
+ "schemars",
+ "serde",
+ "serde_json",
+ "ts-rs",
 ]
 
 [[package]]
@@ -3522,6 +3541,26 @@ dependencies = [
 ]
 
 [[package]]
+name = "ref-cast"
+version = "1.0.26"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "216e8f773d7923bcba9ceb86a86c93cabb3903a11872fc3f138c49630e50b96d"
+dependencies = [
+ "ref-cast-impl",
+]
+
+[[package]]
+name = "ref-cast-impl"
+version = "1.0.26"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "2c9283685feec7d69af75fb0e858d5e7378f33fe4fc699383b2916ab9273e03c"
+dependencies = [
+ "proc-macro2",
+ "quote",
+ "syn 3.0.2",
+]
+
+[[package]]
 name = "regex"
 version = "1.11.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -3875,6 +3914,31 @@ dependencies = [
 ]
 
 [[package]]
+name = "schemars"
+version = "1.2.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "a2b42f36aa1cd011945615b92222f6bf73c599a102a300334cd7f8dbeec726cc"
+dependencies = [
+ "dyn-clone",
+ "ref-cast",
+ "schemars_derive",
+ "serde",
+ "serde_json",
+]
+
+[[package]]
+name = "schemars_derive"
+version = "1.2.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "7d115b50f4aaeea07e79c1912f645c7513d81715d0420f8bc77a18c6260b307f"
+dependencies = [
+ "proc-macro2",
+ "quote",
+ "serde_derive_internals",
+ "syn 2.0.117",
+]
+
+[[package]]
 name = "scoped-tls"
 version = "1.0.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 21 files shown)


tobyhede:
Verified the v3 domain mapping against the actual `eql-bindings` 3.0.1 catalog: all nine term-key combinations it emits (`[]`, `["hm"]`, `["op"]`, `["ob"]`, `["bf"]`, and the unions) map correctly, and storage-only `eql_v3_json` (`Some(&[])` → none) is correctly distinguished from searchable `eql_v3_json_search`/`_entry` (`None` → JsonLike+Contain). The two bug fixes are correct and the regression tests encode the old buggy output as the discriminator — nice.

Findings, most substantive first.

**1. `eql_v3_json_entry` is classified with inverted semantics.** `public.eql_v3_json_entry` is in `v3::all()` and `ste_vec_domain_type!` leaves `term_json_keys` at the default `None`, so `resolve("eql_v3_json_entry")` returns `JsonLike + Contain`. But the installed SQL comments it as `'EQL encrypted JSONB leaf entry (equality, ordering)'` — it's the `->` result element, and its per-entry terms are `hm` XOR `op`. So a column of that type would type-check as supporting `->`/`@>` and *not* `=`/`<`, which is backwards. No user declares such a column today, so it's latent, but `every_public_column_domain_resolves_to_a_known_token_type` actively asserts it resolves. Consider excluding the entry/document operand types from the column catalog rather than relying on nobody declaring them.

**2. `DomainIdentity::canonical` collides with real domains that mean something else.** `canonical(Text, {json_like, ..})` takes the `if traits.json_like { push("search") }` branch and produces `eql_v3_text_search` — which is a *real* catalog domain whose terms are `[hm, op, bf]` (Eq+Ord+TokenMatch), not a JSON capability. This is live: `notes (EQL: JsonLike + Contain)` appears at eight sites in `lib.rs` tests plus `type_env.rs:259`. It can also emit genuinely non-catalog names (`Eq+TokenMatch` → `eql_v3_text_eq_match`, `Contain` → `eql_v3_text_contain`). The doc comment concedes it "is not guaranteed to equal the exact catalog typname" — worth strengthening that to "may collide with an unrelated real domain", or deriving the suffix so it can't produce a real-but-wrong name.

**3. Test-fixture footgun from the new identity field.** `EqlValue` derives `PartialEq`, so identity now participates in equality. Production is safe — both construction sites read traits and identity from the same `ColumnKind::Eql` on the same `Column`, so identity is a function of `TableColumn` and adds no discrimination (and `EqlTerm::Partial`'s union applies to the *bounds* field, not `EqlValue.2`, so nothing desyncs). But fixtures can now disagree: a column declared with the explicit arm `EQL("eql_v3_integer_ord"): Ord` carries identity `eql_v3_integer_ord`, while an expected `projection![(EQL(t.c: Ord))]` builds the canonical `eql_v3_text_ord` — `assert_eq!` fails on a dimension the test wasn't testing. Loud rather than silent, but a papercut for whoever writes the transform tests next.

**4. `docs:` is the wrong commit type.** ~346 added lines are production Rust/SQL, including a new production dependency, a breaking public API change (`ColumnKind::Eql` gains a field, `Column::eql` gains an arg, `ColumnKind` loses `Copy`), a schema-loader rewrite, and a real behaviour fix (`self.eq ^ other.eq` → `self.eq && !other.eq`). The PR's own commits say `feat`/`fix`/`refactor` for three of seven. The body's "designed, not yet built" also undersells it. Matters for changelog generation and bisect.

**5. v2 columns → `native` with only a `warn!`** is right for a v3-only build, and it's not a plaintext-read exposure (a native passthrough returns the raw composite ciphertext). One caveat: `eql_v2.to_encrypted` does no validation and `CREATE CAST (text AS eql_v2_encrypted)` is `AS ASSIGNMENT`, so on the *write* path an explicitly-text-typed, JSON-parseable value (`'42'::text`) could be wrapped and stored in plaintext rather than erroring. Narrow, but it means "fails loudly" isn't quite airtight — worth making that warning hard to miss in ops.

**6. No CHANGELOG entry** — CLAUDE.md asks for one on user-facing changes. Not specific to this PR: none of #422/#423/#424/#426/#427/#428 touch CHANGELOG.md, while the queue collectively changes a public signature, swaps the v2 composite for v3 domains, and bumps cipherstash-client 0.34→0.42. Raising against the queue, not this branch.

tobyhede (changes requested):
Appears to be a working document.<br>Should this be committed?

freshtonic:
@tobyhede I believe it should be - it's valuable context

freshtonic:
Thanks for the thorough review — went through all six findings. Summary of what changed (all pushed):

**1. `eql_v3_json_entry` classified with inverted semantics** — **Fixed.** `eql_v3_json_entry` (the `->` result element, `SteVecEntry`) is now excluded from the column-domain catalog via a `NON_COLUMN_DOMAINS` denylist, so `resolve()` returns `None` for it and a column mistakenly declared with that type falls through to a native column instead of `JsonLike + Contain`. The `every_public_column_domain_resolves_to_a_known_token_type` assertion skips the same list, and a new test pins that it no longer resolves while the real JSON column domains (`eql_v3_json`, `eql_v3_json_search`) still do. → `fix(proxy): exclude eql_v3_json_entry from the column-domain catalog`

**2. `DomainIdentity::canonical` collides with real domains** — **Done** (the doc route you suggested). The doc comment now states the synthesised name may *collide with an unrelated real domain that means something else* (with the `canonical(Text, {json_like}) → eql_v3_text_search` example), and that only `from_domain_name` is authoritative. → `docs(eql-mapper): warn that DomainIdentity::canonical may collide with a real domain`

**3. Test-fixture footgun from the new identity field** — **Acknowledged, no change.** It's loud-not-silent, as you note; leaving it as-is for whoever writes the transform tests next.

**4. `docs:` is the wrong commit type** — **Not reworded.** That's a message rewrite across the stack; leaving it for the merge rather than rebasing the whole queue for the type prefix.

**5. v2 columns → `native` with only a `warn!`** — **Done.** The warning now spells out that the column is served as a **PLAINTEXT (native) column** — Proxy runs no encrypt/decrypt on it, new writes are stored in plaintext, migrate before writing — instead of the softer "ignoring unsupported column". The deeper `AS ASSIGNMENT` write-path concern is inherent to having a v2 column on a v3 build (deliberately native+warn), so I made the warning impossible to miss rather than changing that behaviour. → `feat(proxy): make the v2-column plaintext warning ops-visible`

**6. No CHANGELOG entry** — **Done** (against the queue, per your framing). Added an `[Unreleased]` section covering the whole v3 migration; it lives on #428.

PR #432, opened 2026-07-28
feat: upgrade to EQL v3 — cipherstash-client 0.42, typed jsonb domains, v3 typecheck & transform

freshtonic:
Migrates CipherStash Proxy from EQL v2 to EQL v3.

The central change is how an encrypted column is declared. Under v2 every encrypted column had the same opaque `eql_v2_encrypted` composite type, and its searchable capabilities lived somewhere else entirely — rows in an `eql_v2_configuration` table, populated by `eql_v2.add_search_config` calls. Under v3 the column type *is* the configuration: a typed `jsonb` domain such as `eql_v3_text_search` or `eql_v3_integer_ord` encodes both the scalar type and the searches the column supports.

```sql
-- EQL v2                                  -- EQL v3
encrypted_email eql_v2_encrypted           encrypted_email eql_v3_text_search  -- eq + ord + LIKE
encrypted_dob   eql_v2_encrypted           encrypted_dob   eql_v3_date_ord     -- ord + range + eq
encrypted_salary eql_v2_encrypted          encrypted_salary eql_v3_bigint_ord

SELECT eql_v2.add_search_config(           -- (nothing — the domain says it)
  'users', 'encrypted_email', 'unique', 'text');
SELECT eql_v2.add_search_config(
  'users', 'encrypted_email', 'match', 'text');
-- …and one call per index, per column
```

That single change ripples through the whole stack, which is why this PR is large: the proxy stops reading a config table and infers encrypt config from the schema, the type checker gains a domain identity per column, and every rewrite rule has to emit the v3 functional term form rather than v2's bare operators.

## Breaking changes

- **Existing v2 data and schemas must be migrated to v3.** There is no dual-mode support — this build speaks v3 only. A surviving `eql_v2_encrypted` column is served as a *plaintext* passthrough column with a loud ops-visible warning: Proxy performs no encryption on writes or decryption on reads for it.
- `eql_v2.add_search_config` and `eql_v2.add_encrypted_constraint` are gone, as is the `eql_v2_configuration` table. Declare the domain type instead.
- Requires EQL 3.0.3 installed in the target database.

## Dependencies

| | before | after |
|---|---|---|
| `cipherstash-client` / `cipherstash-config` | 0.34.1-alpha.4 | 0.42.0 |
| `eql-bindings` | — | 3.0.3 |
| EQL (`CS_EQL_VERSION`) | eql-2.3.0-pre.3 | eql-3.0.3 |

Upgrading to `cipherstash-client` 0.42.0 also lets the vendored `stack-auth` copy go — it was the CIP-3159 hotfix patch and is no longer referenced. That deletion accounts for most of the ~12k removed lines; the substantive diff is far smaller than the headline numbers suggest.

## New capabilities

- **Equality on encrypted JSON fields** — `WHERE col -> 'field' = 'value'`, in both the simple and extended protocols, and in the `->>` and `jsonb_path_query_first(col, path) = value` spellings, with `<>` as the negation. The field and the value are fused into a single encrypted value-selector needle and matched by containment, so a query never reveals the field and the value separately. Exact and case-sensitive; the value must be a JSON scalar.
- **Encrypted full-text match with `@@`** on text columns whose domain carries a bloom-filter term.
- **Range comparisons on encrypted JSON fields**, and **array indexing** for encrypted JSON search (`ArrayIndexMode::ALL`).

## Fixed along the way

- `LIKE`/`ILIKE` on an encrypted column is now gated by the column's token-match capability. It previously bypassed capability checking and was silently accepted on columns that cannot do fuzzy match.
- `ORDER BY` on an encrypted column now orders by its ordering term, and `GROUP BY` groups by its equality term, projecting grouped columns through `eql_v3.grouped_value`. Both previously fell through to raw `jsonb` semantics over the ciphertext — which is non-deterministic, so the results were wrong rather than merely unsorted.

## Reviewer's map

The commits are ordered so the story reads top to bottom; reviewing by commit is likely easier than by file.

| area | what to look at |
|---|---|
| `packages/eql-mapper` (+3672/-312) | the bulk. `inference/unifier/types.rs` for `DomainIdentity`/`TokenType`; `transformation_rules/` for the rewrite pipeline — `cast_full_payload_operands`, `rewrite_eql_comparison_ops`, `rewrite_eql_match_ops`, `rewrite_eql_order_by`, `rewrite_eql_group_by`, `rewrite_json_value_selector_eq` are new, replacing `cast_literals_as_encrypted` / `cast_params_as_encrypted` |
| `packages/cipherstash-proxy` (+1684/-804) | `proxy/encrypt_config/from_domain.rs` (config inferred from the domain typname), `proxy/schema/eql_domains.rs`, and the v3 encrypt/decrypt paths in `proxy/zerokms/` and `postgresql/messages/data_row.rs` |
| `docs/` | `docs/plans/2026-07-20-eql-v3-type-checker-handoff.md` is the design record; ADRs cover the domain-identity and rewrite-pipeline decisions |
| `tests/`, `packages/showcase` | fixtures and the healthcare showcase converted to v3 domains |

## Testing

`tests/sql/schema.sql`, the benchmark schema and the showcase schema are all v3 now, and the integration suite is v3-shaped.

Ten backwards-compatibility tests in `packages/cipherstash-proxy-integration/src/eql_regression.rs` are `#[ignore]`d: they pin captured v2 ciphertext and `::eql_v2_encrypted` casts, neither of which exists under v3. They need fixtures regenerated from a v3 baseline before they can be re-enabled.

## Acknowledgment

## Summary by CodeRabbit

* **New Features**
  * Upgraded encrypted SQL processing to EQL v3 using schema-driven, self-configuring encrypted domain types.
  * Enhanced encrypted querying: `@@` fuzzy matching, encrypted JSON equality/inequality with `->`/`->>`, `jsonb_path_query_first`, improved `SELECT DISTINCT` deduping, and `SELECT DISTINCT … ORDER BY` support.
* **Bug Fixes**
  * Correctness improvements for reused parameters, encrypted `GROUP BY`, JSON selector typing, `LIKE`/`ILIKE` capability gating, and prepared/unnamed statement reuse.
* **Documentation**
  * Updated architecture, setup, reference/searchable JSON, errors, and showcase/schema examples for EQL v3.
* **Chores**
  * Updated bundled client/EQL versions; EQL v2-encrypted schemas/data require migration.


packages/eql-mapper/src/type_checked_statement.rs
```diff
@@ -85,14 +137,73 @@ impl<'ast> TypeCheckedStatement<'ast> {
...
+                        value: input,
+                    },
+                    None => OutputParamSource::Input(input),
+                };
+
+                Ok(OutputParam {
+                    param: Param((idx + 1) as u16),
+                    value,
+                    source,
+                    query_operand: self.query_operands.contains_param(input),
```
coderdan: [P1] Track query roles per placeholder occurrence

When one placeholder is reused for storage and querying, e.g. `UPDATE t SET enc = $1 WHERE enc = $1`, this marks both renumbered outputs query-only. The proxy strips ciphertext from the `SET` value, so its stored-domain cast fails. Track the role per AST occurrence/output instead of per input parameter.

packages/cipherstash-proxy/src/postgresql/messages/parse.rs
```diff
@@ -23,19 +23,44 @@ impl Parse {
...
+    /// rewritten SQL, so PostgreSQL can always infer them.
+    pub fn rewrite_param_types(&mut self, output_params: &[OutputParam]) {
+        if self.param_types.is_empty() {
+            return;
+        }
+
+        let param_types = output_params
+            .iter()
+            .map(|output| match &output.column {
+                Some(_) => Type::JSONB.oid() as i32,
```
coderdan: [P2] Declare selector-only parameters as text

When a client supplies Parse OIDs for `json_col -> $1` or `jsonb_path_query(json_col, $1)`, the selector still has an encrypted column configuration, but the rewritten EQL function expects bare `text`. Forcing JSONB here makes PostgreSQL reject the rewritten Parse. Special-case `JsonAccessor`/`JsonPath` outputs as text.

packages/cipherstash-proxy/src/postgresql/data/from_sql.rs
```diff
@@ -93,6 +106,134 @@ pub fn literal_from_sql(
...
+    }
+
+    let value = match param.format_code {
+        FormatCode::Text => {
+            let text = param.to_string();
+            serde_json::from_str::<serde_json::Value>(&text)
+                .unwrap_or(serde_json::Value::String(text))
+        }
+        FormatCode::Binary => {
+            parse_bytes_from_sql::<serde_json::Value>(&param.bytes, postgres_type)
```
coderdan: [P2] Decode binary text JSON-field operands as strings

For extended `->>` equality with a binary TEXT/VARCHAR RHS, normal bytes such as `Alice` are passed directly to `serde_json::Value::from_sql` and rejected instead of receiving the text-format string fallback. The same issue affects `JsonOrd`. Decode textual PostgreSQL types as `String` before JSON parsing.

packages/eql-mapper/src/transformation_rules/rewrite_eql_group_by.rs
```diff
@@ -0,0 +1,185 @@
...
+
+    /// Whether `item` projects one of the grouped encrypted columns, and so must
+    /// be lifted through an aggregate.
+    ///
+    /// Matched on the resolved column rather than on syntax, so `SELECT t.col …
+    /// GROUP BY col` — which PostgreSQL accepts today — keeps working.
+    fn projects_grouped_column(&self, item: &'ast SelectItem, grouped: &[EqlValue]) -> bool {
+        Self::select_item_expr(item)
+            .and_then(|expr| self.eql_value_of(expr))
+            .is_some_and(|value| grouped.contains(&value))
```
coderdan, line 122: [P2] Avoid wrapping aggregate projections in `grouped_value`

For `SELECT MIN(enc) FROM t GROUP BY enc`, `MIN(enc)` resolves to the same `EqlValue` as the grouped column, so this predicate wraps it as `grouped_value(eql_v3.min(enc))`. PostgreSQL rejects that nested aggregate. Only direct grouped-column projections should be lifted.

packages/eql-mapper/src/transformation_rules/rewrite_eql_group_by.rs
```diff
@@ -0,0 +1,185 @@
...
+            }
+            GroupByExpr::All(_) => vec![],
+        }
+    }
+
+    /// The expression a select item projects, if it is a plain one.
+    fn select_item_expr(item: &'ast SelectItem) -> Option<&'ast Expr> {
+        match item {
+            SelectItem::UnnamedExpr(expr) | SelectItem::ExprWithAlias { expr, .. } => Some(expr),
+            _ => None,
```
coderdan, line 97: [P2] Handle wildcard projections in GROUP BY rewrites

For a valid query such as `SELECT * FROM t GROUP BY id, enc`, the encrypted key becomes `eq_term(enc)`, but the wildcard still projects bare `enc`, causing PostgreSQL's “must appear in GROUP BY” error. Expand/lift encrypted wildcard columns or reject this shape before emitting invalid SQL.

coderdan (commented):
Reviewer: Codex (GPT-5.6 Sol).

Five correctness findings follow.

coderdan (changes requested):
Nice work. Findings from the Codex review should be addressed, particularly the P1.

coderdan (changes requested):
Reviewer: **Claude Fable 5** (via Claude Code). This analysis was performed by Claude Fable — a fan-out of five specialised review agents over the diff, with every P1 claim below **empirically verified** by running the actual SQL through the mapper's own test harness (with control queries proving the equivalent supported shapes *are* rewritten). Hygiene is green (fmt, clippy `-D warnings`, compile) and the mapper unit suite passes 116/116.

**Verdict: all five findings from the previous (Codex) review are properly addressed — nice work on `3dfe95e0`, each fix is correct and carries a real reproduction test. However, the review surfaced a new class of confirmed P1 bugs: SQL shapes that type-check but are never rewritten, silently returning wrong results — plus one plaintext-to-database leak.**

## Part 1: Previous review findings — all five addressed ✅

All fixed in `3dfe95e0`, each with a reproduction test:

1. **[P1] Per-occurrence query roles** — Fixed properly. The role now comes from the rewritten AST itself: `RenumberParams` records which output params are wrapped in an `eql_v3.query_*` twin cast, and `param_plan` keys on the output param. The load-bearing assumption checks out end-to-end: the cast emitters (`query_operand_domain` → `cast_expr_to_v3_domain`) and the detector (`is_query_twin`) both use the same two-part `eql_v3.query_*` shape, including the JSON `query_json` special case. Integration tests cover both `SET enc=$1 WHERE enc=$1` and the reversed pairing, asserting the update actually takes effect.
2. **[P2] Selector params declared TEXT** — Fixed in `parse.rs::rewrite_param_types`; `JsonAccessor`/`JsonPath` outputs now declare `TEXT`. Tested via `prepare_typed`, which reproduces the pgx `cache_describe` path.
3. **[P2] Binary text JSON operands** — Fixed with a `binary_json_value` helper that reads textual types as strings first; the `JsonOrd` arm routes through it too. Unit-tested (end-to-end repro impossible since tokio-postgres refuses to serialise the bad bytes — the commit message is honest about this).
4. **[P2] Aggregates not wrapped in `grouped_value`** — Fixed via `is_direct_column_reference`; only bare column refs are lifted. Unit tests cover MIN/MAX plus the still-lifted direct projection.
5. **[P2] Wildcard + GROUP BY rejected** — Fixed with a clear error, consistent with `SELECT DISTINCT *` handling. Tested.

## Part 2: New findings

### P1 — silent wrong results (all empirically confirmed)

The root cause is a **parity gap between the type checker and the rewrite pipeline**: inference accepts shapes the transformer has no rule for, and the fallthrough is "emit unchanged" — which under v3 means raw jsonb comparison against payloads containing the randomised ciphertext `c`.

| Shape | Confirmed output | Silent failure |
|---|---|---|
| `enc IN ('a','b')` / `NOT IN` | payload literals substituted, no `eq_term` wrap | 0 rows / all rows |
| `enc BETWEEN $1 AND $2` | passed through verbatim | arbitrary wrong rows |
| `enc IS DISTINCT FROM $1` | passed through verbatim | equal plaintexts compare distinct |
| `SELECT DISTINCT ON (enc) …` (no ORDER BY) | passed through verbatim | no dedup |
| `SELECT DISTINCT *` | passed through — the wildcard guard in `rewrite_eql_distinct.rs:105` is unreachable when *every* item is a wildcard | no dedup |
| `ORDER BY 1` / `GROUP BY 1` (ordinal → encrypted col) | ordinal untouched | wrong sort order / one group per row |
| `rank() OVER (PARTITION BY enc)` | no rule covers `partition_by` | every row its own partition |
| `UNION` / `INTERSECT` on encrypted columns | no rule for set operations | duplicates survive / empty intersect |

The fix doesn't have to be implementing all of these — rejecting them loudly at type-check time (the pattern this PR uses everywhere else) closes every row in the table. Recommendation: adopt a shared invariant — **anything inference accepts on an EQL type must have a transformation rule, or inference must refuse it.**

**Chained JSON accessor leaks the plaintext selector** — `rewrite_json_value_selector_eq.rs:184` clones the container from the *original* AST. For `profile -> 'a' -> 'b' = $1`, confirmed output is `eql_v3.jsonb_contains(profile -> 'a', $1::…)`: the plaintext selector `'a'` ships in the SQL text and native `->` runs on the encrypted payload. Found independently by two review passes; the clone also reverts any substitution already made inside the container (e.g. an EQL literal in a CASE container reappears as plaintext).

**NULL path param forwards the plaintext value operand to the database** — verified by direct code reading. For `col -> $1 = $2` with `$1` bound NULL, `json_value_selector_plaintext` yields no plaintext, encryption is skipped, and the rebuild path in `bind.rs:172-190` forwards `$2`'s raw client bytes (`apply_encrypted` with `None` is a no-op). The sensitive comparand crosses the wire and can land in the server log when the domain CHECK rejects it. The output param should be rewritten to NULL (SQL semantics: `col->NULL = x` is NULL anyway).

### P2 — should fix

**Mapper:**
- `GROUP BY email ORDER BY email` on an hm-carrying text domain emits `GROUP BY eq_term(email) ORDER BY ord_term(email)` — PostgreSQL rejects with 42803 (confirmed). The tests only use ord-only domains, where the two term functions coincide by accident.
- Valid plaintext-keyed `DISTINCT ON (dept) … ORDER BY dept` is spuriously rejected whenever an encrypted column merely appears in the projection (confirmed), with an error message describing a shape the query doesn't have (`rewrite_eql_distinct_order_by.rs` `applies_to`).
- Reused selector param clobbers fusion: `j->'a' = $1 OR j->'b' = $1` — `by_param` is keyed per input param and `record_param` overwrites, so both needles fuse path `'b'` (`json_value_selector.rs:52`). Same per-occurrence lesson as the earlier P1.
- `j -> $1 = 'x'` (param path, literal value) always hard-fails with `InternalError("param $1 … is not consumed")` — a combination `json_value_selector.rs`'s own docs list as supported.
- When JSON-equality fusion declines (selector is neither literal nor placeholder), the caller still sets `handled = true` and skips the generic binop rule — `j -> some_col = 'secret'` sends the plaintext literal inside `eql_v3.eq_term('secret')` instead of failing the capability check (`expr.rs:168-188`, `641-667`).
- `check_postcondition` is never invoked, so the new "unused encrypted literals" defense in `substitute_encrypted_literals.rs` is dead code — that invariant is exactly what would have caught the chained-accessor bug.

**Proxy:**
- The last commit (`fa5ef700`) introduced a bug: `parse_handler` sets the statement→session mapping at `frontend.rs:755`, then `close_statement` at line 778 deletes it, and nothing re-inserts it (`add_statement` at line 861 only restores the statements map). Every subsequent Bind warns and falls back to "latest session", misattributing metrics under pipelining. Fix: swap the two calls.
- Schema and encrypt-config are loaded by two independent managers running the same query on separate connections and timers; after DDL, a connection can snapshot mismatched halves and then fails with `UnknownColumn` for its entire lifetime. One load should produce both structures.
- Config inference keys by bare `table_name` across all schemas: `audit.users` encrypted + `public.users` plaintext → duplicate-name resolution fails, and with `enable_mapping_errors=false` (the default) both tables silently pass through unmapped.
- v2/missing-EQL databases run in warn-only passthrough. It's not silent (multiple warns, a per-connection banner), and the v3 domain CHECKs prevent plaintext-at-rest — but with the config table gone as a guardrail, consider refusing to start (or requiring an explicit flag) when `eql_v2_encrypted` columns exist or `eql_v3.version()` is absent.

**Tests/docs:**
- CHANGELOG omits two silent behavior changes the PR's own tests document via `#[ignore]`: `jsonb_array_length` now returns 1, and term filters (downcase) are no longer applied — case-insensitive queries on `encrypted_jsonb_filtered` return 0 rows. Boolean equality removal is buried in another entry.
- ARCHITECTURE.md and ADR-0003 both describe a rule chain containing two deleted rules (`CastLiteralsAsEncrypted`, `CastParamsAsEncrypted`) and omit roughly half the shipped chain; `eql-mapper/CONTEXT.md` still claims the capability loader is future work that this PR in fact implements.
- No integration coverage for: the `@@` operator, any of the v2-success→v3-error capability rejections (a regression re-accepting them would pass the suite), or `eql_v3.grouped_value` end-to-end.

### P3 (abridged)

Binary INT4/INT8/FLOAT8/BOOL value operands on JSON fields rejected (driver-dependent gap); fused path param decoded via `Display` ignoring format code; empty-string params conflated with NULL on the widened re-serialization path; new debug logs emit predicate plaintext (`bind.rs:141`, consistent with existing practice but new sites); `from_domain.rs` re-implements domain recognition and disagrees with `eql_domains::resolve` (`eql_v3_json_entry` gets a config entry); wrapped domains/arrays silently degrade to native with no warning; DDL overlay marks new EQL columns `Native` for the connection's lifetime; `docs/errors.md` still recommends v1 `cs_add_column_v1`; the legacy-decrypt regression suite is fully `#[ignore]`d with no v3 baseline fixtures; stale "3.0.2 currently pinned" comment in `rewrite_eql_group_by.rs:44`; misleading literal-count error message in `type_checked_statement.rs:255`.

### Verified clean

The unifier core (no Eql×Native unification path; capability laundering via CAST/COALESCE/functions blocked), `helpers.rs`, `cast_full_payload_operands.rs`, alias preservation, wire-protocol length/count arithmetic on re-serialized messages, error routing (ErrorResponse, not connection-fatal), per-connection statement cache isolation, and the **vendored `stack-auth` removal** — the registry 0.42.0 crate contains both fixes the vendored copy carried (CIP-3233 expiry, CancelGuard refresh), no token logging, no insecure endpoints.

## Part 3: Suggested focus areas for human review

Ranked by leverage — items 1–3 involve judgment calls that automated review can't make:

1. **The typecheck/rewrite parity policy** — Read `type_checked_statement.rs::make_transformer` next to `inference/infer_type_impls/expr.rs` and decide, per unsupported shape (IN, BETWEEN, ordinals, set ops, PARTITION BY, DISTINCT ON): implement or reject-loudly. This is the product decision behind every P1, and it determines whether v3 can ever silently return wrong results. Also decide whether to wire up `check_postcondition` as a hard invariant.
2. **What leaves the proxy: `bind.rs` + `param_plan`** — `rewrite`, `apply_encrypted`, and `json_value_selector_plaintext` are the last gate before the wire. The NULL-fusion plaintext forward lives here; every `None`-ciphertext path should be deliberate. Highest security stakes per line in the PR.
3. **Passthrough posture** — `proxy/mod.rs` startup, the `enable_mapping_errors=false` default, and the v2-database warn-only behavior. Whether "warn and serve" is acceptable now that the schema is the sole config source is a policy call, not a code bug.
4. **The JSON value-selector fusion** — `rewrite_json_value_selector_eq.rs` + `json_value_selector.rs`. Newest and subtlest logic in the PR, three confirmed bugs already; the container-clone-from-original-AST pattern deserves a skeptical read.
5. **Schema→config inference** — `from_domain.rs`, `eql_domains.rs`, and the dual-manager loading. Now the sole source of encryption truth; the cross-schema merge and coherence race both live here.

Reasonable to deprioritise: the `stack-auth` deletion (verified against the registry crate), the unifier core (two independent review passes found it sound), the mechanical test-file migrations (audited — no silent coverage deletions), showcase/benchmark schemas, and docs (beyond fixing the stale tables listed above).

freshtonic:
@coderdan thank you for the extra review. Nearly all the new P2 findings are due to the same error - not wrapping the encrypted column in `eq_v3.<term>(col)` in a bunch of places (it's writing the ciphertext instead)

tobyhede:
Reviewer: **Claude Opus 5** (via Claude Code). Twelve agents, each told to refute rather than confirm, against `edb988d5`. SQL-semantics claims tested on throwaway PostgreSQL clusters with the real pinned EQL v3 and v2.1.8 SQL loaded — which settles the "needs a live EQL install" items left open earlier.


## New

**`SELECT DISTINCT *` emits base-table names — regression from `edb988d5`.** `rewrite_eql_distinct.rs:119-125` builds expanded items from the resolved `TableColumn`, never the alias in `FROM` scope. `SELECT DISTINCT * FROM employees AS e` → `... ON (employees.id, ...) ... FROM employees AS e` → `ERROR: invalid reference to FROM-clause entry for table "employees"`. Same for CTEs, derived tables, self-joins. These got a clean mapper error at `724378b5`. `lib.rs:2963` covers only the unaliased case, and no integration test mentions `DISTINCT`. The same function emits table idents unquoted (`Ident::new` at `schema/manager.rs:138` vs `with_quote` for columns at `:145`).

**`Expr::InList` imposes no capability bound.** `expr.rs:73-85`, against `IsDistinctFrom` (`:66`, `Eq`) and `Between` (`:111`, `Ord`). `WHERE stor IN ('a','b')` type-checks on a zero-capability `eql_v3_text` column where `stor = 'a'` is refused. That bypasses a deliberate gate — `eql_v3_boolean` is storage-only because a two-value column leaks its distribution (`eql_domains.rs:122-123`). `Expr::InSubquery` (`:87-95`) has the same hole and no rewrite rule.

Separately, `IN` on a capable column seq-scans: it lowers to a `ScalarArrayOpExpr`, which blocks the SQL-function inlining that lets `=` match the expression index. Measured on 5000 rows with `enable_seqscan=off` — 4998 rows removed by filter, against an identical `Index Cond` for `BETWEEN`.

**`select_table_schemas.sql` aggregates lack `ORDER BY c.ordinal_position`.** `Table.columns` is an order-sensitive `Vec`, `SELECT *` expands in ordinal order at PostgreSQL, and the decrypt path zips them positionally. Measured reordering with `join_collapse_limit=1`, and before `ANALYZE` the default plan already diverged. Result: encrypted column returned raw while a text column takes its OID — `backend.rs:548` no-ops the mismatch. Pre-existing. (The parallel-`array_agg`-desync mechanism raised on this file separately is not real — measured 0 misaligned across 5 plans including forced parallelism. The ordering fix is still needed.) Also here: `array_agg` doesn't skip NULLs, so `Vec<String>` at `manager.rs:134` panics on a table with no readable columns; and the `table_constraints` join at `:12-15` is unreferenced and a row multiplier.

**A v2-typed column on a v3 build stores plaintext.** EQL v2.1.8 ships `CREATE CAST (text AS eql_v2_encrypted) ... AS ASSIGNMENT`, which bypasses `record_in`. Measured against a replica of the v2 type: any text-typed valid-JSON value stores in plaintext — including `UPDATE` overwriting a real ciphertext row, and `INSERT … SELECT` from a text column. Valid-JSON covers every numeric string. Reads leak the raw envelope unconditionally (`SELECT * FROM v2t` returns undecrypted JSON). On a partially-migrated database `EncryptConfig::is_empty()` is false as soon as one v3 column exists, so the "Data is not protected" warnings never fire.

## Escalations

**Empty-string → NULL (filed P3) corrupts plaintext columns.** Re-serialisation fires when *any* param is dirty (`bind.rs:50-56`), so `INSERT INTO t (enc, plain) VALUES ($1, $2)` with `$2=''` writes NULL into `plain`. Measured round-trip `[-1, 0, 5]` → `[-1, -1, 10]`. With no encrypted param the Bind passes through and `''` survives — the same query corrupts or not depending on an unrelated param. Root cause: `is_null()` is `bytes.is_empty()` (`:266-268`), while the decoder at `:367-377` knows the difference and discards it.

**The chained accessor also reverts already-encrypted literals.** The container clone at `rewrite_json_value_selector_eq.rs:184` discards every rewrite already applied inside that subtree, not just the selector. `WHERE (CASE WHEN name = 'topsecret' THEN profile ELSE profile END) -> 'a' = $1` ships `'topsecret'` — a search value for an encrypted column — in cleartext.

## Confirmed

Verified independently at HEAD; brackets are what I'd add.

NULL path param leak [empty-string path triggers it too] · reused selector param clobbers fusion [the `<>` form is a filter *bypass* — returns rows the predicate excludes] · `DISTINCT ON (enc)` without `ORDER BY` [inference enforces the `Eq` bound the rewrite layer ignores] · set operations [`EXCEPT ALL` broken too; `UNION ALL` safe] · `PARTITION BY` [named windows too; window `ORDER BY` was already handled] · ordinals · plaintext-keyed `DISTINCT ON` rejected [enshrined in `lib.rs:3010`] · `j -> $1 = 'x'` InternalError [three doc comments contradict each other] · `GROUP BY`/`ORDER BY` term mismatch · dead `check_postcondition` · Parse session mapping [diagnostics-only] · dual loader [Context never re-reads; `schema_changed` latch blocks every `ReadyForQuery`] · bare `table_name` keying · binary numeric JSON operands [binary int4 `825373492` is ASCII `"1234"` and parses as JSON `1234`] · DDL overlay [`CREATE TABLE IF NOT EXISTS` on an encrypted table unmaps it for the connection] · fused path via `Display` · plaintext debug logs [gated; but `data/mod.rs:25` writes `target =` not `target:`, so it obeys global level] · wrapped/array domains · `cs_add_column_v1` in docs · CHANGELOG · ADR-0003 + CONTEXT.md [also claims `EqlTraits::all()`, zero hits repo-wide] · `eql_regression.rs` [`fixtures.json` never existed in git history — those tests passed vacuously] · `is_json_entry` [any deserialisation failure ships the envelope to the client] · term filters [scalar match already downcases; loss is SteVec-only] · `jsonb_array_length` · `from_domain.rs` vs `eql_domains::resolve` · stale "3.0.2 pinned" · literal-count message · missing `@@`/`grouped_value`/capability-rejection coverage.

Already fixed: `SELECT * … GROUP BY` (`3dfe95e0`) · `SELECT DISTINCT *` guard (`edb988d5`, see above) · `pg_catalog` restriction · `ARCHITECTURE.md`. The five Codex fixes are all in `3dfe95e0`; the P1 among them has no mapper unit test, only the Docker-gated integration one, and the `JsonOrd` half of finding 3 is untested.

## Didn't reproduce

- `BETWEEN`/`IN`/`IS DISTINCT FROM` returning wrong rows. EQL declares the comparison operators for every domain — capability-matched pairs are real term comparisons ignoring `c`, mismatched pairs are `eql_v3_internal.*` stubs that `RAISE`. Correct or loud, not silent. (The separate `InList` and seq-scan facts above stand.)
- Declined fusion emitting a plaintext literal. `notes -> sel_col = 'secret'` is rejected at type check; reverting `expr.rs` to its pre-`724378b5` form gives byte-identical errors, so it was unreachable in both. Residual: `as_ast_value` (`:721-727`) sees through `Expr::Cast` but not `Expr::Nested`, so `notes -> ('a') = '"x"'` falls through to `eq_term` on a JSON operand — no leak, no overload either.
- `SyntheticNames` colliding with an `__eql_*` schema column — an unprojected column can't reach outer scope.
- Silent `Ok(false)` in `rewrite_eql_distinct_order_by` — replaced all five bail sites with panics: 129 passed, 0 hits. The cited line is already an `Err`.
- The fixture-provenance version bump — `src/` is byte-identical between eql-bindings 3.0.2 and 3.0.3, so the current comment is accurate and the patch would have made it false.

freshtonic:
## Outstanding review feedback, deduplicated

Every finding from the four review passes on this PR — CodeRabbit (5 runs), Codex/GPT-5.6 via @coderdan, Claude Fable 5, Claude Opus 5 — that is **still live at `0532e2bb`**, merged into one list with duplicates collapsed.

Already closed and **not** repeated below:

- **The whole P1 table** from [the Fable review](https://github.com/cipherstash/proxy/pull/432#pullrequestreview-4803351160) — IN/NOT IN, BETWEEN, IS DISTINCT FROM, DISTINCT ON, `SELECT DISTINCT *`, ordinals, PARTITION BY, UNION/INTERSECT — along with the Opus comment's re-confirmations of those same rows, and the capability-bound gaps on `IN` / `IN (subquery)` / `ANY` / `ALL` / `CASE` operands. Fixed across `19944fee`, `08d41a60`, `b4b31201`, `0532e2bb`; each shape is now covered by both an eql-mapper transform test and a proxy integration test. Two caveats that remain live below: the *alias* defect in the `SELECT DISTINCT *` expansion (§2), and the `ALL` set-operation quantifiers (§2).
- All five Codex findings (`3dfe95e0`).
- The CodeRabbit findings fixed in `724378b5` / `edb988d5` — rooted `$` paths, `get_eql_v3_function_name` schema qualification, declined-fusion fall-through, the `Ok(false)`-after-mutation bail-outs, `@@` symmetry, cross-schema config overwrite, synthetic-name collisions, ARCHITECTURE.md, the `->`/`->>` headings, ADR-0003 fences, fixture provenance.

---

### 1. Data integrity and plaintext exposure

**A v2-typed column on a v3 build stores plaintext, and startup won't stop it.**
*(Opus new · CodeRabbit `proxy/mod.rs:88-105` Major · Fable P2 "passthrough posture" — same root, three framings)*
EQL v2.1.8 ships `CREATE CAST (text AS eql_v2_encrypted) … AS ASSIGNMENT`, which bypasses `record_in`. Measured against a replica of the v2 type: any text-typed valid-JSON value stores in plaintext — including `UPDATE` overwriting a real ciphertext row, and `INSERT … SELECT` from a text column. Valid-JSON covers every numeric string. Reads leak the raw envelope unconditionally. On a partially-migrated database `EncryptConfig::is_empty()` is false as soon as one v3 column exists, so the "Data is not protected" warnings never fire. `Proxy::eql_version` still degrades every failure to `None` and only warns on mismatch — there is no minimum-version gate. With the config table gone as a guardrail, consider refusing to start (or requiring an explicit flag) when `eql_v2_encrypted` columns exist or `eql_v3.version()` is absent.

**Empty-string params become NULL, corrupting *plaintext* columns.**
*(Opus escalation of Fable P3)*
`is_null()` is `bytes.is_empty()` (`bind.rs:266-268`), and re-serialisation fires when *any* param is dirty (`bind.rs:50-56`). So `INSERT INTO t (enc, plain) VALUES ($1, $2)` with `$2=''` writes NULL into `plain`. Measured round-trip `[-1, 0, 5]` → `[-1, -1, 10]`. With no encrypted param the Bind passes through and `''` survives — the same query corrupts or not depending on an unrelated param. The decoder at `:367-377` already knows the difference and discards it.

**Chained JSON accessor ships the plaintext selector, and reverts already-encrypted literals.**
*(Fable P1 non-table + Opus escalation; found independently by two passes)*
`rewrite_json_value_selector_eq.rs:184` clones the container from the *original* AST. `profile -> 'a' -> 'b' = $1` emits `eql_v3.jsonb_contains(profile -> 'a', $1::…)` — `'a'` in cleartext, and native `->` running on the encrypted payload. The clone also discards every rewrite already applied inside that subtree: `WHERE (CASE WHEN name = 'topsecret' THEN profile ELSE profile END) -> 'a' = $1` ships `'topsecret'` — a search value for an encrypted column — in cleartext. Test exists, `#[ignore]`d in `19944fee`. Tracked as CIP-3682: the fix is not the one-line container swap it looks like, because containment needs a *composite* path (`$.a.b`) rooted at the bare column, which means composing the path across both inference and the rewrite.

**NULL (or empty-string) selector param forwards the raw value operand to the database.**
*(Fable P1 non-table + Opus confirmed, with the empty-string trigger added)*
For `col -> $1 = $2` with `$1` bound NULL, `json_value_selector_plaintext` yields no plaintext, encryption is skipped, and `bind.rs:172-190` forwards `$2`'s raw client bytes (`apply_encrypted` with `None` is a no-op). The sensitive comparand crosses the wire and can land in the server log when the domain CHECK rejects it. Rewrite the output param to NULL — `col -> NULL = x` is NULL anyway. Test exists, `#[ignore]`d in `19944fee`. Tracked as CIP-3682.

**`select_table_schemas.sql` aggregates lack `ORDER BY c.ordinal_position`.**
*(CodeRabbit ×2 — once Major/security, once nitpick — + Opus new)*
`Table.columns` is an order-sensitive `Vec`, `SELECT *` expands in ordinal order at PostgreSQL, and the decrypt path zips positionally (`schema/manager.rs:140-144`, `encrypt_config/manager.rs:207`). Opus measured reordering with `join_collapse_limit=1`, and before `ANALYZE` the default plan already diverged. Result: an encrypted column returned raw while a text column takes its OID — `backend.rs:548` no-ops the mismatch. Pre-existing, still unfixed. Two more on the same file: `array_agg` doesn't skip NULLs, so `Vec<String>` at `schema/manager.rs:134` panics on a table with no readable columns; and the `table_constraints` join (lines 12-15) is unreferenced and a row multiplier.

### 2. Mapper

- **`SELECT DISTINCT *` emits base-table names — regression from `edb988d5`.** *(Opus new)* `rewrite_eql_distinct.rs:119-125` builds expanded items from the resolved `TableColumn`, never the alias in `FROM` scope. `SELECT DISTINCT * FROM employees AS e` → `… ON (employees.id, …) … FROM employees AS e` → `ERROR: invalid reference to FROM-clause entry for table "employees"`. Same for CTEs, derived tables, self-joins. These got a clean mapper error at `724378b5`. `lib.rs:2963` covers only the unaliased case, and no integration test mentions `DISTINCT`. The same function emits table idents unquoted (`Ident::new`, `schema/manager.rs:138`) where columns use `with_quote` at `:145`.
- **`GROUP BY email ORDER BY email` on an hm-carrying text domain** emits `GROUP BY eq_term(email) ORDER BY ord_term(email)` — PostgreSQL rejects with 42803. The tests only use ord-only domains, where the two term functions coincide by accident. *(Fable P2, Opus confirmed)*
- **Valid plaintext-keyed `DISTINCT ON (dept) … ORDER BY dept` is spuriously rejected** whenever an encrypted column merely appears in the projection, with an error message describing a shape the query doesn't have (`rewrite_eql_distinct_order_by.rs::applies_to`). Enshrined in `lib.rs:3010`. *(Fable P2, Opus confirmed)*
- **Reused selector param clobbers fusion.** `j->'a' = $1 OR j->'b' = $1` — `by_param` is keyed per input param and `record_param` overwrites, so both needles fuse path `'b'` (`json_value_selector.rs:52`). Opus adds the sharp edge: the `<>` form is a filter **bypass**, returning rows the predicate excludes. Same per-occurrence lesson as the Codex P1. *(Fable P2 + Opus)*
- **`j -> $1 = 'x'`** (param path, literal value) always hard-fails with `InternalError("param $1 … is not consumed")` — a combination `json_value_selector.rs`'s own docs list as supported. Opus notes three doc comments contradict each other on this. *(Fable P2 + Opus)*
- **`check_postcondition` is never invoked**, so the "unused encrypted literals" defence in `substitute_encrypted_literals.rs:64` is dead code. Confirmed at HEAD — the only caller is the tuple impl in `mod.rs:194`, which nothing drives. That invariant is exactly what would have caught the chained-accessor leak above. *(Fable P2, Opus confirmed)*
- **Residual on the `724378b5` fusion-decline fix:** `as_ast_value` (`expr.rs:721-727`) sees through `Expr::Cast` but not `Expr::Nested`, so `notes -> ('a') = '"x"'` falls through to `eq_term` on a JSON operand — no leak, but no overload either. *(Opus)*
- **`IN` on a capable column seq-scans.** It lowers to a `ScalarArrayOpExpr`, which blocks the SQL-function inlining that lets `=` match the expression index. Measured on 5000 rows with `enable_seqscan=off`: 4998 rows removed by filter, against an identical `Index Cond` for `BETWEEN`. Separate from the correctness question, which is settled. *(Opus new)*
- **`EXCEPT ALL` / `INTERSECT ALL` are still broken** (`UNION ALL` is genuinely safe). Confirmed against the refusal that landed in `08d41a60`: `deduplicates()` keys on the quantifier and lets `SetQuantifier::All` through by design, but `EXCEPT ALL` and `INTERSECT ALL` still match rows across branches by equality, so they compare whole payloads including the randomised `c`. Only `UNION ALL` needs no equality at all. *(Opus, refining the excluded table row)*
- **Length guard before the projection `zip`** in `rewrite_eql_distinct_order_by.rs:259-287` — if any earlier rule ever changes projection length, the zip silently drops trailing columns. Distinct from the `Ok(false)` bail-outs fixed in `724378b5`. *(CodeRabbit nitpick)*

### 3. Proxy

- **`fa5ef700` regression, still present:** `parse_handler` sets the statement→session mapping at `frontend.rs:755`, `close_statement` at `:778` deletes it, and nothing re-inserts it (`add_statement` at `:861` only restores the statements map). Every subsequent Bind warns and falls back to "latest session", misattributing metrics under pipelining. Fix: swap the two calls. Diagnostics-only. *(Fable P2, Opus confirmed)*
- **Dual loader.** Schema and encrypt-config are loaded by two independent managers running the same query on separate connections and timers; after DDL a connection can snapshot mismatched halves and then fail with `UnknownColumn` for its entire lifetime. Opus adds: `Context` never re-reads, and the `schema_changed` latch blocks every `ReadyForQuery`. One load should produce both structures. *(Fable P2 + Opus)*
- **Bare `table_name` keying — partially addressed.** `edb988d5` restricted the scan to `current_schemas(false)` and keeps the first match, but both maps are still keyed on the bare name, so two search-path schemas holding the same table still collide; with `enable_mapping_errors=false` (the default) both silently pass through unmapped. The SQL's own comment documents the remaining exposure. *(Fable P2 + CodeRabbit `encrypt_config/manager.rs:225` + Opus)*
- **`from_domain.rs` re-implements domain recognition and disagrees with the catalog.** `from_domain.rs:40` decides JSON searchability by `domain.ends_with("_search")`; `eql_domains.rs` derives it from eql-bindings via `term_json_keys() == None`. A future searchable JSON domain not ending in `_search` would type-check as `JsonLike + Contain` yet be configured storage-only, so writes would silently carry no SteVec terms. Also gives `eql_v3_json_entry` a config entry it shouldn't have. Confirmed still divergent. *(CodeRabbit nitpick + Fable P3 + Opus)*
- **`is_json_entry`:** any deserialisation failure ships the envelope to the client. *(Fable P3, Opus confirmed)*
- **Binary INT4/INT8/FLOAT8/BOOL value operands on JSON fields rejected** (driver-dependent gap). Opus adds the trap: binary int4 `825373492` is ASCII `"1234"` and parses as JSON `1234`. *(Fable P3 + Opus)*
- **Fused-path param decoded via `Display`,** ignoring the format code. *(Fable P3, Opus confirmed)*
- **New debug logs emit predicate plaintext** (`bind.rs:141`) — consistent with existing practice, but new sites. Opus found one that is *not* gated: `data/mod.rs:25` writes `target = MAPPER` rather than `target: MAPPER`, so it obeys the global level instead of the MAPPER target. Confirmed at HEAD. *(Fable P3 + Opus)*
- **DDL overlay marks new EQL columns `Native` for the connection's lifetime.** Opus's concrete case: `CREATE TABLE IF NOT EXISTS` on an already-encrypted table unmaps it for that connection. *(Fable P3 + Opus)*
- **Wrapped domains and arrays silently degrade to native with no warning.** *(Fable P3, Opus confirmed)*
- **`UnsupportedV3QueryTerm` and `UnsupportedSteVecInV2` both collapse to `InvalidIndexTerm`**, whose `#[error("InvalidIndexTerm")]` (`error.rs:254`) hands the customer an opaque identifier — unlike the new SteVec variants. `UnsupportedV3QueryTerm` is reachable from a real query shape. Repo guideline: customer-facing errors carry a friendly message and a doc link. *(CodeRabbit nitpick)*
- **`statement.rs:130-132`:** add a debug length assertion that `get_output_param_columns()` matches `plan.outputs()` before the positional `zip`, so a future mismatch is explicit rather than a silently dropped trailing output. *(CodeRabbit nitpick)*
- **`schema/manager.rs:172-174`:** the legacy-column warning fires once per column per reload tick for the process lifetime. Rate-limit, metricise, or emit only on schema change. *(CodeRabbit nitpick)*

### 4. Tests and coverage

- **`eql_regression.rs` is fully `#[ignore]`d, and the stated plan can't unblock it.** `insert_encrypted_directly` still casts fixture payloads to `eql_v2_encrypted`, so v3-generated fixtures fail reinsertion. Ten sites: `157` (generation), `268` text, `300` int2, `330` int4, `360` int8, `390` float8, `420` boolean, `450` JSONB decrypt, `480` JSON field access, `532` JSON array. Opus adds the fact that changes the priority: **`fixtures.json` never existed in git history — those tests were passing vacuously.** Still one of only two unresolved threads on the PR. *(CodeRabbit Major + Opus)*
- **No integration coverage** for the `@@` operator — a regression re-accepting it would pass the suite. `eql_v3.grouped_value` is now exercised end-to-end by the `GROUP BY` tests in `operator_class_shapes.rs`. The v2-success→v3-error capability rejections are deliberately mapper-only: a capability error is not observable through the proxy while mapping errors fall back to passthrough, so an integration test would assert the opposite of what it looks like it asserts (CIP-3680). *(Fable P2, Opus confirmed)*
- **`indexing.rs:38`** — `let _ = client.simple_query(sql).await` swallows a failed `CREATE INDEX`, so the test passes even if `eql_v3.ord_term` is the wrong term function for `eql_v3_text_search`, which is the one thing the v3 index migration needs to prove here. Tolerate only "already exists". Confirmed still `let _ =`. *(CodeRabbit nitpick)*
- **The Codex P1 fix has no mapper unit test** — only the Docker-gated integration one — and the `JsonOrd` half of Codex finding 3 is untested. *(Opus)*
- **`from_sql.rs`** — table-driven unit tests for `json_selector_path`, `json_value_selector_plaintext`, `literal_json_value`, `json_ord_scalar_plaintext`: rooting, scalar-only rejection, invalid-JSON-as-string fallback, number→f64. Several JSON integration tests are ignored under v3, so this is the only place that behaviour gets pinned. *(CodeRabbit nitpick)*
- **`renumber_params.rs:80-100`** — pin the `<placeholder>::JSONB::eql_v3.query_*` cast shape, so a future change to `cast_expr_to_v3_domain` fails loudly instead of making `wrapped_placeholder` silently return `None`. *(CodeRabbit nitpick)*
- **`lib.rs:2601-2655`** — the DISTINCT/ORDER BY table covers neither `LIMIT` retention on the wrapper nor the `?column?` fallback from `derive_effective_alias`. *(CodeRabbit nitpick)*
- **`psql-passthrough.sh`** lost its only EQL-specific assertion when the `eql_v2_configuration` probe went; it now checks `SELECT 1` twice. Add `SELECT eql_v3.version();` through the proxy. Confirmed. *(CodeRabbit nitpick)*
- **`test_error_messages.py:78`** — the second unresolved thread. `724378b5` added the `finally`; the follow-up is to run the cleanup DELETE only after the transaction context exits (an aborted transaction can't clean up) and to use a collision-resistant id. *(CodeRabbit nitpick)*

### 5. Docs and changelog

- **CHANGELOG omits two silent behaviour changes** that the PR's own `#[ignore]`d tests document: `jsonb_array_length` now returns 1, and term filters (downcase) are no longer applied, so case-insensitive queries on `encrypted_jsonb_filtered` return 0 rows. Opus refines the second: scalar match already downcases, so the loss is SteVec-only. Boolean-equality removal is buried inside another entry. *(Fable P2 + Opus)*
- **ADR-0003 and `eql-mapper/CONTEXT.md` are still stale.** ARCHITECTURE.md was fixed in `724378b5`; ADR-0003 only got fence languages, and still describes a chain containing the deleted `CastLiteralsAsEncrypted`/`CastParamsAsEncrypted` while omitting roughly half the shipped chain. `CONTEXT.md` still calls the capability loader future work that this PR in fact implements. Both also claim `EqlTraits::all()`, which has zero hits repo-wide. *(Fable P2 + Opus)*
- **`docs/errors.md` still recommends v1 `cs_add_column_v1`.** *(Fable P3, Opus confirmed)*
- **Stale "3.0.2 currently pinned"** at `rewrite_eql_group_by.rs:44`. Confirmed. *(Fable P3 + Opus)*
- **Misleading literal-count error message** at `type_checked_statement.rs:255`. *(Fable P3, Opus confirmed)*
- **`model/schema.rs:51`** — `#[display("Eql({})", _0)]` renders only the `EqlTraits`, so the `DomainIdentity` the v3 loader populates never appears in diagnostics, which is exactly the field you'd want when debugging a domain mismatch. Confirmed. *(CodeRabbit nitpick)*

---

### Raised but disproved — recording these so they don't come back

Opus tested these on throwaway PostgreSQL clusters with the real pinned EQL v3 and v2.1.8 SQL loaded, and could not reproduce them:

- **`BETWEEN` / `IN` / `IS DISTINCT FROM` returning wrong rows.** EQL declares the comparison operators for every domain: capability-matched pairs are real term comparisons ignoring `c`, mismatched pairs are `eql_v3_internal.*` stubs that `RAISE`. Correct or loud, never silent. `19944fee` reached the same conclusion independently by test.
- **Declined fusion emitting a plaintext literal.** `notes -> sel_col = 'secret'` is rejected at type check; reverting `expr.rs` to its pre-`724378b5` form gives byte-identical errors, so the path was unreachable in both.
- **`SyntheticNames` colliding with an `__eql_*` schema column** — an unprojected column can't reach outer scope.
- **Silent `Ok(false)` in `rewrite_eql_distinct_order_by`** — all five bail sites replaced with panics: 129 passed, 0 hits. The cited line is already an `Err`.
- **Parallel `array_agg` desync in `select_table_schemas`** — 0 misaligned across 5 plans including forced parallelism. (The *ordering* fix in §1 is still needed; it's a different mechanism.)
- **The fixture-provenance version bump** — `src/` is byte-identical between eql-bindings 3.0.2 and 3.0.3, so `724378b5`'s resolution was right and the proposed patch would have made the comment false.

---

### The one structural decision behind most of this

Both Fable and Opus land on the same recommendation, and it's worth deciding explicitly rather than case by case: **anything inference accepts on an EQL type must have a transformation rule, or inference must refuse it.** The fallthrough today is "emit unchanged", which under v3 means a raw jsonb comparison against payloads containing the randomised ciphertext `c` — silently wrong rather than loud. Wiring `check_postcondition` up as a hard invariant is the enforcement half of the same decision.

coderdan (approved):

PR #440, opened 2026-08-04
fix(mapper): close inference gaps on AST fields of type-checked statements (CIP-3699)

freshtonic:
Closes the fail-open gaps from [CIP-3699](https://linear.app/cipherstash/issue/CIP-3699): statements that pass `requires_type_check` had clauses that escaped inference entirely, so encrypted columns in them were **silently wrong** — plaintext stored unencrypted, or operations executing on raw ciphertext — instead of loudly rejected. Every gap is now either fully supported (inference bound + transformation rewrite) or explicitly rejected.

### 1. `INSERT … ON CONFLICT DO UPDATE SET` — supported
The common upsert path. Previously the `DO UPDATE` assignments were never resolved, so `ON CONFLICT (id) DO UPDATE SET enc = 'plaintext'` stored plaintext in the encrypted column whenever the conflict fired.

- Assignments are now resolved against the target table (mirroring plain `UPDATE … SET`), so literals and params on the conflict path get the column's EQL type and are encrypted, with the existing `Assignment` cast rule applying the full-payload domain cast.
- The `excluded` pseudo-table is brought into scope with the target's projection, so `SET enc = excluded.enc` and `excluded.<col>` in the `DO UPDATE … WHERE` predicate carry the column's encrypted type; predicate comparisons are rewritten to search terms by the existing rules.
- The insert target relation is now **named** (alias, else table name) so `t.col` resolves in `RETURNING` / `DO UPDATE`. Unqualified column references in `DO UPDATE` expressions become ambiguous (target vs `excluded`) — mirroring PostgreSQL's own `column reference is ambiguous` error there.
- **Rejected**: a conflict target naming an encrypted column (`ON CONFLICT (enc)`) — a conflict only fires off a unique index, and uniqueness of the randomised ciphertext never collides, so the upsert would silently always insert. `ON CONFLICT ON CONSTRAINT <name>` is let through: the mapper has no constraint catalog to resolve the name against, and a constraint over an encrypted column is a schema-design error independent of any statement.

Where: `packages/eql-mapper/src/inference/infer_type_impls/insert_statement.rs`, `packages/eql-mapper/src/importer.rs`.

### 2. Window `ORDER BY` and frames — supported (frames partially rejected)
`row_number() OVER (ORDER BY enc)` previously ordered on ciphertext. The window's `ORDER BY` keys now get an `Ord` bound (the `OrderByExpr` rewrite already fired inside window specs, so the missing piece was the capability check). Frame offsets are unified as native. `RANGE` frames **with an offset** over an encrypted sort key are rejected — offset `RANGE` needs `key ± offset` arithmetic, which no term supports; `ROWS` and `GROUPS` frames work (the ordering term is deterministic, so peer groups are preserved).

Where: new `packages/eql-mapper/src/inference/infer_type_impls/window_spec.rs`.

### 3. Named windows (`OVER w` / `WINDOW w AS (…)`) — supported
Inference and rewriting moved from the `Function` node to the `WindowSpec` node itself, which exists in both the inline `OVER (…)` form and the named `WINDOW w AS (…)` definition. Named windows now get the same `Eq` bound + `eq_term` rewrite on `PARTITION BY` and `Ord` bound on `ORDER BY` as inline specs, checked once at the definition site; `OVER w` needs nothing of its own. The `rewrite_eql_partition_by.rs` skip of `NamedWindow` is gone (the rule now matches `WindowSpec` nodes directly), so the transform no longer contradicts inference.

Where: `window_spec.rs`, `packages/eql-mapper/src/transformation_rules/rewrite_eql_partition_by.rs`, `function.rs`.

### 4. Aggregate argument clauses — mixed
- `count(DISTINCT enc)` — **supported**: `Eq` bound on `DISTINCT` arguments, and a new rule rewrites the argument to `eql_v3.eq_term(enc)`. The equality term is deterministic per plaintext, so distinct terms = distinct plaintexts; previously the dedup ran on randomised payloads and the count silently equalled the row count. Sound for `count` precisely because `count` discards its argument values.
- Any **other** aggregate with a `DISTINCT` encrypted argument — **rejected** at transform: the same substitution would change that aggregate's result (`min(DISTINCT enc)` would return the term, not the payload). Note: functions outside the EQL registry already force native arguments, so in practice this bites `min`/`max`.
- `array_agg(x ORDER BY enc)` — **supported**: `Ord` bound; the existing `OrderByExpr` rewrite produces `ORDER BY eql_v3.ord_term(enc)`.
- `WITHIN GROUP (ORDER BY enc)` — **rejected**: ordered-set aggregates compute their result *from* the sort key, so a term rewrite would hand the client an opaque, undecryptable term. Native sort keys are unified as native (same closure as unknown-function arguments), so nothing escapes as an unresolved variable.

Where: `function.rs`, new `packages/eql-mapper/src/transformation_rules/rewrite_eql_aggregate_distinct.rs`.

### 5. `SELECT … INTO` — rejected when encrypted
The projection lands in a table the encryption schema has never seen — unreachable ciphertext. Rejected when any projected column (including through wildcards) is encrypted; native-only `SELECT INTO` passes through unchanged.

Where: `select.rs`.

### 6. `ORDER BY ALL` — rejected
The PostgreSQL dialect never parses it (DuckDB/ClickHouse syntax), so this is defense in depth on the AST shape: the `OrderByKind::All` arm now errors instead of silently matching nothing.

Where: `query_statement.rs`.

### Also closed while in there
`MySQL INSERT … SET` (the `Insert.assignments` field), `ON DUPLICATE KEY UPDATE`, and unknown variants of the non-exhaustive `OnInsert` enum are rejected explicitly instead of ignored.

## Notes for review

- **`Ord` implies `Eq`** in `EqlTraits` (equality falls back to the ordering term), so the `Eq`-bound rejection tests use a storage-only domain (`eql_v3_boolean`).
- Naming the insert target relation is a small behavioural widening for all `INSERT`s (qualified `t.col` now resolves where it previously errored); no existing tests changed behaviour.
- Two existing named-window tests (`window_function_with_forward_reference`, `common_table_expressions`) used a capability-less `salary (EQL)` column under `window w AS (… order by salary …)` — they only passed because the clause escaped inference. They now grant `Ord`.

## Testing

- `mise run check` (fmt, clippy `-D warnings`, compile): **pass**
- `cargo test -p eql-mapper`: **pass** — 154 tests (20 new: transformed-SQL assertions for every supported shape, explicit-error assertions for every rejected shape)
- `mise run test:unit`: 277/278 pass; the one failure is `proxy::tests::init_zerokms_client_with_crn`, which **also fails on a clean checkout of `origin/main`** in this environment (env-dependent ZeroKMS init test) — pre-existing, unrelated.
- `mise run test:integration`: **not run** — the suite tears down and rebuilds the shared Postgres containers and needs CipherStash credentials not available in the isolated worktree this was built in. Please treat integration coverage of the upsert path as unverified.

CIP-3699


freshtonic (commented):
Comment-only review.\n\nI did not find a blocking issue in the diff. The newly covered AST fields are either given explicit capability/type constraints and rewrites or rejected, and the tests cover the supported and rejected cases. The full PostgreSQL 14–17 matrix and performance check are green.

tobyhede (approved):

PR #441, opened 2026-08-04
fix(proxy): protocol desync on transformation error after a mapped statement (CIP-3678)

freshtonic:
Fixes [CIP-3678](https://linear.app/cipherstash/issue/CIP-3678).

## The bug

Once a connection had successfully run a mapped (encrypted) statement, any later statement failing during transformation surfaced to the client as a protocol error — tokio_postgres `UnexpectedMessage` — instead of the proxy's actual error, and the proxy's useful message was never delivered. With connection pools, mapped-then-failing is the common case.

## Diagnosis

A wire trace of the reproduction (protocol-level logging on both sides of the proxy) shows the desync is a **response-stream reorder**, not a client- or state-machine bug:

1. After `query()` returns, tokio_postgres drops its cached prepared statement and pipelines `Close(s0)` + `Sync`, immediately followed by the failing statement's `Parse(s1)`, `Describe`, `Sync`.
2. The frontend failed the Parse locally and wrote `ErrorResponse` + `ReadyForQuery` **straight to the client**, while the server's `CloseComplete` + `ReadyForQuery` for the Close were still in flight.
3. The client received `E, Z, CloseComplete, Z` where the protocol requires `CloseComplete, Z, E, Z`. It attributed the error to the Close request and the Close's responses to the failed prepare — `UnexpectedMessage`.

On a fresh connection there is nothing in flight, so the direct reply happened to line up — which is why the bug only appeared after a prior statement. The ticket's stale-session hypothesis was a real (smaller) leak — the failed Parse's session-metrics entry was abandoned — but it was not the cause of the desync.

## The fix

The frontend no longer replies to the client out of band. A failed Query/Parse/Bind is **replaced with a `DO ... RAISE EXCEPTION` statement sent to the server**, carrying the proxy's message, SQLSTATE, hints and related fields via `RAISE ... USING`. Because the injected statement is a request like any other, the server emits the `ErrorResponse` and `ReadyForQuery` in exactly the slot the client expects — after any in-flight responses, with accurate transaction status. The rest of the failed batch, including its `Sync`, is discarded (the injected statement already produced the batch's `ReadyForQuery`).

This also drains the session the failed statement had started, through the backend's normal ErrorResponse completion path, instead of leaking it.

Two deliberate carve-outs:

- **FATAL errors** (e.g. `SET CIPHERSTASH.KEYSET_ID` with a configured default keyset) are still written directly to the client: `RAISE` cannot produce FATAL severity, and ordering is moot because the client abandons the connection — and everything pipelined on it — on receipt. `ErrorState` is now an enum making the two delivery modes and their differing Sync handling explicit.
- The exception body is nested through `quote_literal` at both levels (no dollar quoting), so error text containing quotes, `%` or `$$` cannot escape the statement. The previous (dormant) `to_database_exception` helper had both hazards.

Also fixed along the way: `parse_handler` recorded the statement's session mapping and then immediately wiped it via `close_statement` (which drops the mapping for the rebound name). Every Bind then hit the "Session lookup failed for prepared statement, using latest session" fallback. The close now happens before the mapping is recorded.

## Regression tests

`extended_protocol_error_messages::tests`:

- `transformation_error_after_mapped_statement` — mapped warm-up, then `SELECT id FROM encrypted WHERE encrypted_bool = $1` (storage-only column, no equality term). Asserts the error arrives as a proper `db error` containing the proxy's message, and that the connection remains usable afterwards. Fails with `UnexpectedMessage` against the unfixed proxy; passes with the fix.
- `transformation_error_after_passthrough_statement` — same failing statement on a passthrough-warmed connection (this desynced locally too, timing-dependent; now deterministic either way).

Also adds `common::proxy_port()` (`CS_PROXY__PORT`, default 6432) so a local run can target a proxy on a non-default port.

## Test evidence

Run locally against a proxy built from this branch (port 6533, `CS_DEVELOPMENT__ENABLE_MAPPING_ERRORS=true`, shared dev postgres on 5532):

- `mise run check` — clean (fmt, clippy, compile).
- `mise run test:unit` — all green (258 result lines, 0 failures).
- Both new regression tests: fail (`UnexpectedMessage`) against the unfixed frontend, pass with the fix.
- Integration slice (`extended_protocol_error_messages`, `set_keyset_error`, `simple_protocol`, `pipeline`, `passthrough`, `map_params`, 34 tests): 27 pass. The 7 failures were re-run against an **unfixed** build of the same branch and fail identically there — pre-existing in this environment, not regressions:
  - `multitenant::ore_order::*` (3) — require a proxy with no default keyset (CI runs them in a separate phase).
  - `passthrough_{invalid_statement,insert_from_select,select_with_cardinality}` and `encrypted_column_not_defined_in_schema` — expect type-check failures to fall through to the database; in this environment (shared dev DB, EQL 3.0.3 installed vs 3.0.4 built) the mapper rejects them before the DB does, on both builds.
- Not run: the full integration suite and the Python/Go language suites (the standard port 6432 was occupied by an unrelated process on this machine, so everything above ran against a dedicated instance on 6533; the language suites hardcode 6432). CI should exercise these.

## Notes for review

- Client-visible change on non-FATAL proxy errors: the ErrorResponse now originates from PostgreSQL's `RAISE`, so `where`/`file`/`line`/`routine` fields point at the `inline_code_block` rather than being absent, and a `Position` field is no longer set (parse errors already embed line/column in the message). Message text, SQLSTATE, severity, hint, table/column fields are preserved via `RAISE ... USING`.
- A failed statement inside an open transaction now genuinely aborts the server-side transaction (matching vanilla PostgreSQL), where previously the proxy refused client-side and left the server transaction open while the client believed it had failed.
- The injected simple-protocol `Query` destroys the server's unnamed prepared statement/portal (protocol-defined side effect). A client that re-binds a previously parsed unnamed statement across a failed batch would need to re-Parse; this was also the proxy's behaviour before the direct-write error path was introduced (ab60e961, Feb 2025).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
 ### Fixed
 
+- **Statement errors no longer desync the connection**: when a statement failed inside the proxy (an unsupported operation on an encrypted column, for instance), the error was written straight to the client and could overtake responses still in flight from the server — with connection pools and prepared-statement caching, the client then saw a protocol error (`unexpected message from server` in tokio_postgres) instead of the proxy's message, typically right after an encrypted statement had run on the same connection. The proxy now delivers statement errors through the server, so clients always receive the proxy's actual error message, in order, and the connection remains usable.
+
 - **A param bound as both a stored value and a query operand**: `UPDATE t SET enc = $1 WHERE enc = $1` failed with a domain CHECK violation. The two occurrences need different payloads — the stored one carries the ciphertext, the query one only search terms — but the role was tracked per input param, so marking the param as a query operand stripped the ciphertext from the value being stored. The role is now taken from the rewritten statement, per occurrence.
 
 - **JSON selector params when the client declares its own types**: a client that sends param OIDs in Parse (pgx in `cache_describe` mode, for example) got `function eql_v3.jsonb_path_exists(eql_v3_json_search, jsonb) does not exist`. A JSON field selector is passed to the rewritten function as bare text, but was being declared as `jsonb` like every other encrypted operand. Affects `->`, `->>`, `jsonb_path_exists`, `jsonb_path_query` and `jsonb_path_query_first`.
--- a/packages/cipherstash-proxy-integration/src/common.rs
+++ b/packages/cipherstash-proxy-integration/src/common.rs
@@ -47,6 +47,17 @@ use tracing::info;
 use tracing_subscriber::{filter::Directive, EnvFilter, FmtSubscriber};
 
 pub const PROXY: u16 = 6432;
+
+/// Proxy port for tests: `CS_PROXY__PORT` if set, otherwise [`PROXY`].
+///
+/// Lets a local run target a proxy on a non-default port without patching the
+/// test source (mirrors [`get_database_port`] for `CS_DATABASE__PORT`).
+pub fn proxy_port() -> u16 {
+    std::env::var("CS_PROXY__PORT")
+        .ok()
+        .and_then(|s| s.parse().ok())
+        .unwrap_or(PROXY)
+}
 pub const PROXY_METRICS_PORT: u16 = 9930;
 pub const PG_PORT: u16 = 5532;
 pub const PG_TLS_PORT: u16 = 5617;
--- a/packages/cipherstash-proxy-integration/src/extended_protocol_error_messages.rs
+++ b/packages/cipherstash-proxy-integration/src/extended_protocol_error_messages.rs
@@ -2,7 +2,13 @@
 mod tests {
     use tracing::{debug, info};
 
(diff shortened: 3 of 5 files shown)


freshtonic (commented):
Comment-only review.\n\nThe ordering strategy—injecting the error through PostgreSQL and distinguishing injected exceptions from directly-sent fatal errors—makes sense, and the close-before-session-recording fix is important.\n\nFinding: the two regression tests added by this PR,  and , currently fail in every PostgreSQL matrix job (14–17). Because those tests directly exercise the behavior this PR is intended to fix, this is not a check failure I would treat as incidental; the test/runtime mismatch needs investigation before merge.

freshtonic (commented):
Correction to my preceding review (shell formatting stripped identifiers): the failing regression tests are transformation_error_after_mapped_statement and transformation_error_after_passthrough_statement.

tobyhede (approved):
Don't have enough context for this one.

PR #456, opened 2026-08-18
fix(deps): bump cipherstash-client, cipherstash-config, and cts-common to 0.42.2

tobyhede:
Pulls in stack-auth 0.42.2, which bumps jsonwebtoken from 9.3.1 to 10.4.0.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -729,9 +729,9 @@ dependencies = [
 
 [[package]]
 name = "cipherstash-client"
-version = "0.42.0"
+version = "0.42.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "6c8714a2997ab5a8cc2c871f01dcae9ff7c6b342ed51c5c0d7c8edd4d954c9d1"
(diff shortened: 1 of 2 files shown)


freshtonic (approved):
Approving.

`Cargo.toml` and `Cargo.lock` are consistent, and the full test matrix (PostgreSQL 14–17) plus the performance regression check are green.

The part of this worth stating explicitly, since the PR body describes it only as a version bump: `jsonwebtoken` 9.3.1 → 10.4.0 swaps the crypto backend from `ring` to `aws-lc-rs` (`stack-auth` 0.42.2 requests `default-features = false, features = ["aws_lc_rs", "use_pem"]`). For a service that builds release/Docker images this is the kind of change that can break a build without touching a test — but it's safe here: `aws-lc-rs`/`aws-lc-sys` are already in the lock on `main` via `rustls`, so the build environment already carries the C/CMake toolchain and nothing new is introduced.

Also carried along, both consistent with the 0.42.2 line: `zeroize` 1.8.1 → 1.9.0 (plus `zeroize_derive` 1.4.2 → 1.5.0) and `zerokms-protocol` 0.12.26 → 0.12.28.

PR #458, opened 2026-08-19
feat(proxy): accept inbound EQL ciphertext payloads

freshtonic:
## Summary

- accept EQL v3 storage payloads and query-only operands in SQL literals and bind parameters before plaintext decoding
- validate payload version, inferred destination, active keyset metadata, configured scalar and SteVec SEM terms, authenticated descriptors, and SteVec array markers
- authenticate every stored ciphertext, including every SteVec entry, then independently rederive and compare its searchable metadata; ordinary SELECT decryption remains root-only
- accept bare selector hashes using the fixed selector-hash shape and document the unavoidable plaintext/hash ambiguity
- forward valid payloads without re-encryption and fail closed with a generic, documented transaction-aborting error
- cover application-generated payloads, query-only operands, selector hashes, metadata and term tampering, accepted SteVec storage/readback, non-root SteVec tampering, and transaction recovery in `cipherstash-proxy-integration`
- demonstrate storage, query-only, and selector-hash inputs in the executable showcase

## Compatibility

On every encrypted column type, Proxy reserves three JSON object shapes for application-generated EQL:

- storage payloads with top-level `v` and `i` plus at least one of `c`, `h`, or `sv`;
- scalar query payloads with `v` and `i` plus at least one of `hm`, `bf`, `ob`, or `op`; and
- SteVec query payloads whose only top-level key is `sv`.

Matching values are validated as EQL rather than encrypted as plaintext, with no opt-out. Query-only shapes are rejected in storage positions. The [error documentation](https://github.com/cipherstash/proxy/blob/main/docs/errors.md#encrypt-invalid-inbound-eql-payload) includes a complete upgrade audit query and guidance for text-backed source data.

This PR also restores the configured-default-keyset behavior: when a connection has not selected a keyset, Proxy now scopes encryption and decryption to `CS_DEFAULT_KEYSET_ID` instead of silently using the ZeroKMS credential's account default. Operators whose two defaults differ must decrypt affected existing values with the old account-default keyset and re-encrypt them under the configured default.

## Testing

- `mise run test:check`
- `cargo nextest run -p cipherstash-proxy --no-fail-fast` (151 passed)
- `mise run test:clippy`
- `mise run test:format`
- `cargo test -p cipherstash-proxy-integration --no-run`

Runtime integration and showcase cases use the existing credentialed CI tasks; this checkout lacks Docker access and the application/ZeroKMS credentials needed to start an isolated current-branch Proxy locally.

Linear: CIP-3822


packages/cipherstash-proxy/src/postgresql/inbound_eql.rs
```diff
@@ -0,0 +1,159 @@
...
+
+    let ciphertext: EqlCiphertext =
+        serde_json::from_value(value).map_err(|_| EncryptError::InvalidInboundCiphertext)?;
+    validate_metadata(&ciphertext, column)?;
+    Ok(Some(ciphertext))
+}
+
+fn validate_metadata(ciphertext: &EqlCiphertext, column: &Column) -> Result<(), EncryptError> {
+    if ciphertext.version() != EQL_SCHEMA_VERSION_V3
+        || ciphertext.identifier() != &column.identifier
```
tobyhede, line 99: **Ciphertext metadata is not authenticated (High) — Path A: destination relabeling**

The AEAD tag (the cryptographic seal that proves a ciphertext was not changed) covers only the `c` (ciphertext) and `descriptor` fields. It does not cover the `i` (identifier) field checked here. This check confirms the identifier matches the destination column, but it does not confirm the identifier is genuine — the field is self-reported, not cryptographically bound.

**Exploit path:** An attacker with direct database access (not through Proxy) reads raw ciphertext bytes from a source column. The attacker changes the `i` field to name a different column, and sends the value through Proxy to that new column. Proxy accepts it if the new column uses the same index-term profile (see the companion comment on `validate_scalar_terms` below). Decryption succeeds, because the AEAD tag does not check the destination column name at decrypt time either. Any user with ordinary SELECT access to the new column, through Proxy, can then read the source plaintext. This crosses a privilege boundary: direct database access becomes ordinary Proxy access.

**Recommendation:** Bind the identifier field into the AAD (the authenticated data that seals the AEAD tag), so a relabeled identifier fails decryption rather than passing a structural check.

packages/cipherstash-proxy/src/postgresql/inbound_eql.rs
```diff
@@ -0,0 +1,159 @@
...
+        match index.index_type {
+            IndexType::Unique { .. } => hmac = true,
+            IndexType::Match { .. } => bloom = true,
+            IndexType::Ore => ore = true,
+            IndexType::Ope => ope = true,
+            IndexType::SteVec { .. } => return Err(EncryptError::InvalidInboundCiphertext),
+        }
+    }
+
+    if payload.hmac_256.is_some() != hmac
```
tobyhede: **Ciphertext metadata is not authenticated (High) — Path B: term splicing**

Same root cause as the comment on line 33: the AEAD tag does not cover the `hm`, `bf`, `ob`, `op` term fields checked here. This check confirms the expected term slots are populated for the column's configured indexes, but it does not confirm the terms were derived from the same plaintext as `c`.

**Exploit path:** An attacker with standard write access to this new feature — the access this PR grants — encrypts two values, X and Y, on the same column through Proxy. The attacker copies the `hm`/`bf`/`ob`/`op` fields from the X ciphertext onto the Y ciphertext, and sends the spliced value through Proxy. Proxy accepts it, because every expected slot is populated. The row decrypts to Y but stays indexed and matched as X. This corrupts equality, ordering, and search results with no error signal. Unlike the destination-relabeling path, this needs no special access — normal use of the new feature is enough.

**Recommendation:** Bind the term fields into the AAD, or have Proxy independently re-derive and verify the terms from the decrypted plaintext before accepting an inbound payload.

tobyhede (commented):
Reviewed the new inbound EQL ciphertext path. One security issue with two exploit paths, one design trade-off to confirm, and one operational gap. Each item was checked against the actual code on this branch and the `cipherstash-client` 0.42.2 dependency source before being posted here. Two items from an earlier automated pass did not hold up under verification and are not included: the `"h"` top-level key check is not dead code (it detects SteVec payloads, which carry no top-level `c` field), and the binary-parameter byte handling in `bind.rs` does not corrupt data (original bytes stay untouched, and a failed trial parse falls back safely).

See inline comments for details.

packages/cipherstash-proxy/src/postgresql/inbound_eql.rs
```diff
@@ -0,0 +1,159 @@
...
+/// Ordinary JSON remains plaintext; malformed payload-shaped JSON fails closed.
+pub fn parse(bytes: &[u8], column: &Column) -> Result<Option<EqlCiphertext>, EncryptError> {
+    let Ok(value) = serde_json::from_slice::<Value>(bytes) else {
+        return Ok(None);
+    };
+    let Some(object) = value.as_object() else {
+        return Ok(None);
+    };
+
+    let payload_shaped = object.contains_key("c")
```
tobyhede: **Fail-closed rejection of plaintext values with a `c` key (Medium — confirm intent)**

Proxy treats any JSON value with a top-level `c` key as a ciphertext-shaped payload. This looks deliberate — it is documented in the module comment above, and a test asserts the fail-closed behavior. But the check has a real cost: plaintext JSON with a short top-level key — `c` for "code", "comment", or "content" — fails with a generic error instead of normal encryption.

**Recommendation:** Confirm this trade-off is intended, and call it out in user-facing docs for the feature. Consider whether the error message can hint at the cause without creating a decryption oracle (a signal an attacker could use to test guesses against a ciphertext).

packages/cipherstash-proxy/src/postgresql/frontend.rs
```diff
@@ -1205,6 +1228,37 @@ where
...
+        }
+
+        let ciphertexts = positions
+            .iter()
+            .map(|(_, ciphertext)| Some(ciphertext.clone()))
+            .collect();
+        self.context
+            .decrypt(ciphertexts)
+            .await
+            .map_err(|_| EncryptError::InvalidInboundCiphertext)?;
```
tobyhede: **Decrypt failures are not logged before the error is discarded (Low)**

`authenticate_and_merge_inbound` maps every `context.decrypt` failure to the same generic `InvalidInboundCiphertext` error via `map_err(|_| ...)`. The underlying cause — a tamper attempt, bad key material, a missing key, or a transient ZeroKMS error — is discarded here, before any log statement runs downstream.

During a ZeroKMS outage, this failure looks identical, in the logs, to a tamper attempt. On-call staff cannot tell the two apart from server logs alone.

**Recommendation:** Log the underlying error, at debug or warn level, at this call site before mapping it to the generic client-facing error. Keep the generic error in the client response, to avoid an oracle.

packages/cipherstash-proxy/src/postgresql/inbound_eql.rs
freshtonic: Addressed in 0f51ee07. Inbound payloads now require the AEAD-authenticated record or key-header descriptor to equal the canonical inferred destination (table/column), in addition to the structural identifier check. Relabeling the identifier therefore cannot move a ciphertext across columns. The integration regression encrypts for some_other_table/encrypted_text, rewrites only the claimed table, and verifies Proxy rejects it with the generic error. The showcase and docs demonstrate the canonical descriptor requirement.

packages/cipherstash-proxy/src/postgresql/inbound_eql.rs
freshtonic: Addressed in 0f51ee07. After AEAD authentication, Proxy independently re-encrypts the recovered plaintext for the inferred destination and compares the complete query projection, including every scalar or SteVec SEM term, against the inbound payload. Any mismatch fails closed. Added unit coverage plus an end-to-end regression that splices the equality, match, and ordering terms from X onto Y and verifies rejection.

packages/cipherstash-proxy/src/postgresql/inbound_eql.rs
freshtonic: Addressed in 0f51ee07 by removing the compatibility cost. Detection now requires the combined EQL advertisement shape: version, identifier, and at least one storage field (ciphertext, key header, or SteVec). Ordinary plaintext JSON with a top-level c field remains plaintext and has explicit unit coverage; malformed values that advertise the full EQL shape still fail closed.

packages/cipherstash-proxy/src/postgresql/frontend.rs
freshtonic: Addressed in 0f51ee07. The underlying decrypt or authentication error is now logged at warn level under the ENCRYPT target before it is mapped to the generic client response. Re-encryption and metadata-verification failures are logged similarly, while SEM mismatches get a generic server-side warning without logging attacker-controlled payload details.

packages/cipherstash-proxy/src/postgresql/middleware/frontend.rs
```diff
@@ -1205,6 +1228,84 @@ where
...
+            .collect();
+        let plaintexts = self.context.decrypt(ciphertexts).await.map_err(|err| {
+            warn!(
+                target: ENCRYPT,
+                client_id = self.context.client_id,
+                msg = "Inbound EQL ciphertext authentication failed",
+                error = ?err,
+            );
+            EncryptError::InvalidInboundCiphertext
+        })?;
```
freshtonic: Decrypt does the authentication

packages/cipherstash-proxy/src/postgresql/inbound_eql.rs
tobyhede: Commit 0f51ee07 narrows the check. It now looks for three keys together: `v`, `i`, and one of `c`, `h`, `sv`. This lowers the risk. It does not remove the risk.

Example: the plaintext value `{"v": 2, "i": "order-123", "c": "some code"}` still matches the shape. The parser still fails to decode it. The write still fails.

The test `malformed_payload_shape_fails_closed` (inbound_eql.rs:164) confirms this. It uses the same three keys and shows the fail-closed result.

Two points stay open:

1. No setting turns the check off. Users cannot opt out, per column or globally.
2. The check runs on every column already configured for EQL encryption — text-cast and json/jsonb-cast alike. It does not touch plaintext columns.

Recommendation: Document this trade-off before release. State which key combination triggers the check. Give users a way to scan existing data for the pattern before upgrade.

```
Marked as addressed —
the gap just wears a narrower
coat. Still lets rain in.
```

freshtonic: Addressed in 4db9e51a. The changelog, PR description, and customer-facing error documentation now state the exact advertisement shape (`v` + `i` + any of `c`/`h`/`sv`), that it applies on encrypted columns with no opt-out, and the compatibility consequence. `docs/errors.md` also includes a PostgreSQL audit query users can run before upgrading, plus guidance for text-backed source data.

tobyhede (commented):
## Review summary

This review covers eight candidate findings. Two were not correct, and this review removes them. Six remain.

The security work in this PR is good. The proxy now decrypts every SteVec entry. This finds forged ciphertext, changed entry order, duplicate entries, and swapped entry ciphertext. The comments below do not change that result.

---

### 1. High — the new decrypt path also changes every SELECT

`packages/cipherstash-proxy/src/proxy/zerokms/zerokms.rs:413`

The `decrypt` function now makes one ZeroKMS record for each SteVec entry. Before this PR it made one record for the root entry only.

`EncryptionService::decrypt` has two callers:

- `postgresql/frontend.rs:1273` — `merge_inbound_eql`, the new inbound path.
- `postgresql/backend.rs:497` — the `DataRow` flush, the usual SELECT path.

The change therefore applies to every read of an encrypted JSONB column, not only to inbound payloads.

cipherstash-client makes two entries for each JSON node. A document with 50 nodes gives approximately 100 records. `MessageBuffer` holds 4096 rows. One SELECT of 4096 rows can thus send approximately 409,600 records in place of 4,096.

The client divides the keys into groups of 500, and sends 5 requests at the same time. There is no hard limit to reach. But the number of requests increases from approximately 9 to approximately 820. Each entry also gets its own key-retrieval payload, and the client does not remove the duplicates.

This is not only a cost. The AAD contains the selector. Decryption of the entries on read therefore finds a graft of value entries in the database. That is a true safety gain.

Confirm the intent:

- If the read-path change is intentional, add a CHANGELOG entry. Give the read cost for JSON columns.
- If it is not intentional, expand the entries only in `merge_inbound_eql`. The trait has one real implementation and two test mocks. The change is small.

No test covers the read path for this behaviour.

---

### 2. Medium — the keyset change has no CHANGELOG entry

`packages/cipherstash-proxy/src/proxy/zerokms/zerokms.rs:172`

`init_cipher` now uses the configured `default_keyset_id` when the connection sets no keyset. Before this PR it passed `None`. `None` selects the default keyset of the client credential.

The history shows a repair, not a new decision. Commit `8df6bee7` passed the configured id. Commit `6cbb7119` removed it on 2025-08-13. The releases v2.1.20 to v3.0.1 contain that regression. This PR puts the original behaviour back.

The effect on users is small, because `stash setup --proxy` makes the two ids the same. But if the two ids are different, these things occur after the upgrade:

- Rows written before the upgrade do not decrypt.
- Search terms for old rows do not agree with new terms.
- A wrong `default_keyset_id` now fails at cipher init with `UnknownKeysetIdentifier`. Before, the proxy ignored it.

Only the commit message records this change. Add a CHANGELOG entry and a short migration note.

---

### 3. Medium — two rejected JSON shapes have no documentation and no audit query

`packages/cipherstash-proxy/src/postgresql/inbound_eql.rs:57`

The proxy rejects three JSON shapes in a storage position:

- `v` and `i` and one of `c`, `h`, `sv` — documented.
- `v` and `i` and one of `hm`, `bf`, `ob`, `op` — not documented.
- `sv` as the only key — not documented.

The audit query in `docs/errors.md` needs `v` and `i`:

```sql
WHERE value ? 'v' AND value ? 'i' AND value ?| ARRAY['c', 'h', 'sv']
```

The document `{"sv": ["a", "b"]}` has no `v` and no `i`. The query does not find it, but the proxy rejects it. An operator can run the audit, get zero rows, upgrade, and then get a write failure.

`{"sv": ...}` alone is a possible document. `sv` is the ISO-639 code for Swedish.

`parse` also runs before any check of the column type. See `frontend.rs:659` and `bind.rs:157`. This applies to all encrypted columns, not only to jsonb columns. An encrypted text column that receives `'{"sv":"Hej"}'` also fails.

Change the audit query to find all three shapes. Say in the compatibility note that the reservation includes the query-operand shapes, and that it applies to every encrypted column type.

---

### 4. Low — the SteVec check does not compare the `a` marker

`packages/cipherstash-proxy/src/postgresql/inbound_eql.rs:195`

For two SteVec payloads, `sem_terms_match` does not use the fast path. It compares `into_query_operand()` on the two sides.

In cipherstash-client 0.42.2, `into_query_operand()` maps `SteVecEntryV3` to `SteVecQueryEntryV3 { selector, term }`. It removes `is_array`. The doc comment on that function states this.

The AAD is the descriptor, the key tag, and the 16-byte selector. The encrypted plaintext is the node value. `is_array` is in neither. The `a` field is therefore not authenticated and not compared. The proxy stores it as it arrives.

An attacker who holds a legitimate payload can change `a` and send the payload again. In EQL 3.0.4, `a` controls `eql_v3_internal.is_ste_vec_array`. That function controls `eql_v3.jsonb_array_length` and `eql_v3.jsonb_array_elements`. EQL Mapper sends the standard SQL functions to these. A changed `a` makes one of these results:

- An error becomes a wrong answer with one element.
- A correct answer becomes an error.

There is no effect on the keys or on the plaintext. The severity is low. But the fix is easy: compare `is_array`, or compare the `sv` array directly.

The doc comment at `inbound_eql.rs:190` is also not correct. It says the comparison keeps "every scalar or SteVec SEM term". It does not keep `a`.

---

### 5. Low — the inbound `keyset_id` has no check

`packages/cipherstash-proxy/src/postgresql/inbound_eql.rs:94`

`validate_storage_metadata` checks the version, the identifier, the descriptor, and the term shape. It does not check `EncryptedRecord.keyset_id`. It does not check `KeyHeader.keyset_id` for SteVec payloads. The AAD contains neither field.

This is not a failure of keyset scope. `ScopedCipher::decrypt` always gives an explicit keyset, and that keyset replaces the value in the record. See `vitur_client/mod.rs:785`. The proxy does not use `decrypt_fallible`. A payload that is truly encrypted under a different keyset fails the AEAD check.

But a false value stays in the record. The client groups the payloads by `keyset_id` before it replaces the value. Rows with many different false values therefore divide one key-retrieval request into many requests. The false value also goes into the stored data and into the logs. An external tool that decrypts the same column without an explicit keyset fails.

Add an equality check against the cipher keyset, or set the field to the cipher keyset. This is defence in depth.

---

### 6. Low — `literals_to_plaintext` has no callers

`packages/cipherstash-proxy/src/postgresql/frontend.rs:1524`

Line 668 now calls `literals_to_plaintext_skipping`. Nothing calls `literals_to_plaintext`. The function is a simple call to the `_skipping` version with an empty slice. The behaviour of the two functions is the same.

CI does not fail. `packages/cipherstash-proxy/src/lib.rs:1` has `#![allow(dead_code)]`. This hides the function from clippy `-D warnings`.

Remove the function. Also examine the crate-wide `allow(dead_code)`. It hides all dead code from CI.

---

## Items checked and cleared

These items looked wrong, but they are correct. No action is necessary.

**The containment cast to `eql_v3.query_json` does not break a GIN index.**

`rewrite_containment_ops.rs:121` changes the cast, and the test comment about GIN index use is gone. The change is safe:

- `cipherstash-encrypt.sql` has no GIN operator class. The only operator class is a B-tree class for ORE.
- The rewrite makes a function call to `eql_v3.jsonb_contains`. That function takes plain `jsonb` on both sides. Both domains are domains over `jsonb`.
- The documented GIN expressions apply to the column, not to the needle. This PR does not change the column side.
- The old test asserted a cast to a text domain on a JSON column. That cast cannot work against a real database.
- EQL 3.0.4 names `$1::jsonb::eql_v3.query_json` as the correct needle shape.

**Query-only operands must carry all configured terms. This is correct.**

`validate_query_metadata` uses `validate_scalar_term_presence` without a change. This agrees with the `eql_v3.query_*` domain CHECK, and PostgreSQL rejects a short operand in any case. cipherstash-client documents `into_query_operand()` as the way to make a multi-term operand. The tests use a column with unique, ope, and match indexes, and the tests pass.

---

## Test coverage gap

No integration test stores an inbound SteVec document with success. The tests cover text storage, query-only scalar operands, query-only SteVec needles, bare selectors, and SteVec rejection. They do not cover an accepted inbound SteVec store.

The SteVec comparison uses positional equality of a JSON array. The result therefore depends on agreement between the entry order from the application and the entry order from the re-encryption of the proxy. No test proves that these two orders agree.

Add a test that stores an inbound SteVec document and reads it back.

---

## Not checked

`EncryptedRecord.decryption_policy` comes from the same untrusted source, and it goes to `retrieve_key_payload`. This review did not establish whether the server-generated ZeroKMS MAC makes that field safe.

packages/cipherstash-proxy/src/postgresql/inbound_eql.rs
```diff
@@ -0,0 +1,532 @@
...
+        EqlCiphertext::SteVec(payload) => {
+            if !has_ste_vec_terms(column) || payload.ste_vec.is_empty() {
+                return Err(EncryptError::InvalidInboundCiphertext);
+            }
+            Ok(())
+        }
+    }
+}
+
+fn validate_query_metadata(query: &EqlQueryPayload, column: &Column) -> Result<(), EncryptError> {
```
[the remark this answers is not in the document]
freshtonic: Addressed in 7b58470e. Added direct negative coverage for a JsonOrd operand with stray scalar terms, a scalar operand on an unsupported EQL term, and an empty SteVec operand. Also added direct coverage for scalar storage payloads against SteVec columns and for stored inbound EQL with no destination column.

packages/cipherstash-proxy/src/postgresql/inbound_eql.rs
```diff
@@ -0,0 +1,532 @@
...
+        .iter()
+        .any(|index| matches!(index.index_type, IndexType::SteVec { .. }))
+}
+
+/// Compare all searchable metadata after the plaintext has been authenticated
+/// and independently re-encrypted for the inferred destination column.
+/// `into_query_operand` removes only record ciphertext/key material, leaving
+/// the identifier and every scalar or SteVec SEM term. Bloom-filter positions
+/// are compared without regard to order; all other terms compare exactly.
+pub fn sem_terms_match(inbound: &EqlCiphertext, derived: EqlCiphertext) -> bool {
```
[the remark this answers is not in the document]
freshtonic: Addressed in 7b58470e. Added the end-to-end rejects_ste_vec_terms_spliced_from_another_plaintext regression, which grafts an ordering term from one application-generated JSON payload onto another and verifies the insert fails closed. SteVec comparison now explicitly checks version, identifier, entry count/order, selector, array marker, and every entry term.

packages/cipherstash-proxy/src/postgresql/inbound_eql.rs
```diff
@@ -0,0 +1,532 @@
...
+    match (
+        serde_json::to_value(inbound.clone().into_query_operand()),
+        serde_json::to_value(derived.into_query_operand()),
+    ) {
+        (Ok(inbound), Ok(derived)) => inbound == derived,
+        _ => false,
+    }
+}
+
+fn bloom_filters_match(inbound: &Option<Vec<i16>>, derived: &Option<Vec<i16>>) -> bool {
```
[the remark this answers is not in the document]
freshtonic: Addressed in 7b58470e. Added bloom_filter_presence_mismatch_does_not_match, covering both the presence-sensitive branch and the fail-closed result through sem_terms_match.

freshtonic:
Addressed the remaining review feedback in `7b58470e`:

- scoped all-entry SteVec authentication to inbound storage validation; ordinary `DataRow`/SELECT decryption remains root-only, with direct regression coverage for both modes;
- documented the restored `CS_DEFAULT_KEYSET_ID` behavior and migration implications;
- expanded the compatibility note and audit SQL to cover storage, scalar-query, and bare-SteVec query shapes on every encrypted column type;
- compare SteVec `a`/array metadata during independent derivation and reject tampering end to end;
- require embedded scalar and SteVec keyset metadata to match the active cipher keyset;
- removed the new dead `literals_to_plaintext` wrapper. I also checked the crate-wide `allow(dead_code)`: removing it exposes 11 pre-existing warnings across unrelated CLI, connection, diagnostics, context, and data helpers, so this focused change leaves that existing policy intact;
- added successful inbound SteVec storage/readback integration coverage plus the requested SteVec term-splice, Bloom presence-mismatch, negative query-shape, scalar-vs-SteVec, and missing-destination tests;
- renamed the shared error to `InvalidInboundEqlPayload`, since query-only payloads contain no ciphertext;
- confirmed `decryption_policy` is authenticated by ZeroKMS's tag-version-1 policy MAC during key retrieval and documented that invariant in the decrypt path.

The PR description now carries the full compatibility and migration notes. Static checks, formatting, Clippy, and all 151 proxy tests pass locally; the credentialed PostgreSQL matrix is running in CI.

tobyhede:
## Review

Head `7b58470e`. Three low-priority points. No defect found.

### 1. Low — SteVec order safety depends on a dependency feature flag

`packages/cipherstash-proxy/src/postgresql/inbound_eql.rs:205`

`sem_terms_match` compares SteVec entries by position. This is correct today, on one condition: the `preserve_order` feature of `serde_json` stays off. With that feature off, `serde_json::Map` is a `BTreeMap`, so key order is sorted. The client walk is a sequential depth-first traversal (`json_indexer/path_values.rs:90`). Both sides encode the plaintext through `Value` (`plaintext/mod.rs:332`, `:428`), which removes the key order of the original text. Equal documents therefore give equal entry order.

Two events break the condition, and neither changes this file:

- A new dependency turns on `serde_json/preserve_order`. Cargo unifies features across the graph, so Proxy gets `IndexMap` order.
- A client in another language makes the payload and its core turns on that feature. `napi` supplies such a feature. That client sends insertion order, Proxy derives sorted order, and Proxy rejects every document from it.

The second event matters most, because this feature exists to accept payloads that other SDKs make.

Options: record the condition in a comment, or sort entries by selector before the comparison. The length check already limits the cost.

### 2. Low — the exact-term rule for query operands is not documented

`docs/errors.md:365`

The "How to fix" text gives the exact-term rule for storage payloads only. Query operands obey the same rule: the terms must equal the configured indexes of the column. The error is deliberately generic, so a user who debugs a rejected operand has no other source. Add one sentence.

### 3. Low — the `sem_terms_match` doc comment omits `k`

`packages/cipherstash-proxy/src/postgresql/inbound_eql.rs:190`

The comment lists what the comparison covers, but does not mention `k`. State that the type constrains `k` to one value, and that the comparison therefore excludes it.

tobyhede (approved):
Nice work

PR #464, opened 2026-09-06
fix(deps): patch rand to 0.9.3

tobyhede:
Dependency patch for rand to 0.9.3.

This is an automated security patch update.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -864,7 +864,7 @@ dependencies = [
  "postgres-protocol",
  "postgres-types",
  "proptest",
- "rand 0.9.2",
+ "rand 0.9.3",
  "recipher 0.1.3",
  "regex",
  "rust_decimal",
@@ -912,7 +912,7 @@ dependencies = [
  "fake 4.2.0",
  "hex",
  "postgres-types",
- "rand 0.9.2",
+ "rand 0.9.3",
  "reqwest",
  "rustls",
  "serde",
@@ -1721,7 +1721,7 @@ dependencies = [
  "chrono",
  "deunicode",
  "dummy 0.11.0",
- "rand 0.9.2",
+ "rand 0.9.3",
 ]
 
 [[package]]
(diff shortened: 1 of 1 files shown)


freshtonic (changes requested):
The rand remediation is correct and CI is green: the vulnerable 0.9.2 lock entry is replaced by the patched 0.9.3 release, with no affected rand version left in the lockfile.\n\nBlocking: commit fcc941ff has no `Signed-off-by:` trailer. CONTRIBUTING.md requires every commit to be signed off under the DCO. Please amend the commit with `--signoff` and force-push.\n\nNon-blocking: the tempfile entry also moves from getrandom 0.3.2 to 0.4.2. That appears to be harmless unrelated lockfile resolution churn and could be minimized, but it is not grounds for rejection.

freshtonic (approved):
Re-review complete: the prior DCO blocker is fixed. Commit `9a0bf22e` now has the required `Signed-off-by` trailer and retains a valid Conventional Commit subject.

The CIP-3961 remediation remains correct: the vulnerable `rand 0.9.2` lock entry is replaced by patched `0.9.3`, all dependency edges resolve to it, and the other locked rand versions (`0.8.6` and `0.10.2`) are also patched for GHSA-cq8v-f236-94qc. The lockfile resolves successfully with `--locked --offline`.

The unrelated `tempfile`/`getrandom` resolution change remains harmless and non-blocking. Approved.

freshtonic (approved):
Approved on re-review of 9a0bf22e. The required DCO sign-off is present, and the lockfile exactly reproduces the rand 0.9.2 to 0.9.3 remediation for GHSA-cq8v-f236-94qc. The remaining rand entries (0.8.6, 0.9.3, and 0.10.2) are outside the advisory ranges. The tempfile edge selecting the already-present getrandom 0.4.2 is valid resolver output and non-blocking.

Non-blocking documentation note: AGENTS.md asks notable changes to update CHANGELOG.md Unreleased, but this security patch has no changelog entry. That documentation omission does not block this approval.

Local all-target workspace checks pass; GitHub CI is still in progress.

PR #465, opened 2026-09-18
fix(deps): bump tokio-postgres to 0.7.18

tobyhede:
## Summary

Bumps `tokio-postgres` in `Cargo.lock` from **0.7.13 → 0.7.18** to address [GHSA-3gjw-f78c-vvpw](https://github.com/advisories/GHSA-3gjw-f78c-vvpw) (Medium; affected `>= 0.4.0, < 0.7.18`). Resolves Dependabot alert #55.

Lockfile-only change; no `Cargo.toml` edits (all manifests already use `tokio-postgres = "0.7"`).

### Lockfile changes
- `tokio-postgres` 0.7.13 → 0.7.18
- Companion/transitive bumps pulled in by the resolver: `postgres-types` 0.2.9 → 0.2.14, `postgres-derive` 0.4.6 → 0.4.9, `whoami` 1.6.0 → 2.1.3, `wasite` 0.1.0 → 1.0.2, `phf`/`phf_shared` 0.11.3 → 0.13.1, `serde_json` 1.0.140 → 1.0.151; added `objc2-core-foundation`, `objc2-system-configuration`, `zmij`
- `libredox` 0.1.3 → 0.1.24 (required: `whoami` 2.x needs a newer `libredox` than was locked)
- `postgres-protocol` (0.6.12) and `tokio-postgres-rustls` (0.13.0) unchanged

## Verification
- `cargo tree -i tokio-postgres --workspace`: single version, `tokio-postgres v0.7.18`
- `cargo check --workspace --all-targets`: passes


CHANGELOG.md
```diff
@@ -20,6 +20,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
 ### Fixed
 
+- **Database error detail in Proxy's logs**: logged database errors again include the message PostgreSQL returned, not just the error kind. The upgraded PostgreSQL client library stopped appending the underlying cause when an error is rendered as text, which left entries such as `Database connection error` reading only `db error`. Proxy now renders the cause itself, so the server's message is back in the log line.
```
freshtonic: I think this entry describes a regression that no user saw. The `tokio-postgres` bump and the `ErrorChain` fix are both in this PR, so a released Proxy never logged a bare `db error`. 3.0.1 logs `db error: ERROR: …` and so does this branch. A reader of the release notes would think that database errors in the logs were broken in a released version.

I suggest that you remove this entry, or merge one sentence into the `### Security` entry below. For example: "Logged database errors still include the message from PostgreSQL." Not blocking.

freshtonic (approved):
Approving. The dependency bump is correct and the follow-up work is sound.

**What I checked**

- `ErrorChain` gives the same log text as before the bump. Under 0.7.13, `Display` rendered `db error: ERROR: <msg>`. Now `ErrorChain` renders the kind (`db error`), then `: `, then the `DbError` source (`ERROR: <msg>`). `ConfigError::Database` and `Error::Config` are `#[error(transparent)]`, so both `Display` and `source()` forward to the inner error and the chain stays correct through the crate wrappers.
- No client-facing path sends a `tokio_postgres::Error` to the client. The only `tokio-postgres` clients are the schema loader, the EQL version probe and the `migrate` CLI. Their errors are logged, not converted in `error_to_response`, so no client can get a bare `db error` in an `ErrorResponse`.
- The remaining `err.to_string()` log sites (`ConnectionTimeout`, prometheus, TLS, config reload) cannot contain a `tokio_postgres::Error`, so it is correct to leave them unchanged.
- `assert_db_error` compares `severity()` and `message()` on the `DbError`. This is more precise than the old string compare, and the passthrough test no longer needs its own nested `match`.
- All required checks pass on `9eaa4bc`.

**Non-blocking**

1. The PR description is out of date. It says "Lockfile-only change; no `Cargo.toml` edits", but the branch now also adds `ErrorChain`, changes nine log sites, adds two test helpers and adds CHANGELOG entries. Please update the summary so the squash commit and the release notes match the change.
2. See the inline comment on the `### Fixed` CHANGELOG entry.
3. Small point: `ErrorChain` skips a cause when the text rendered so far ends with that cause. A short cause (for example `timeout`) can match the end of an unrelated message and get dropped. This is acceptable for log output. If you want it exact, compare only against the parent's own `Display`, not the accumulated string. No change is necessary.

